PR #150, opened 2015-01-31
Add support for versioned shareable objects for amd64.Linux.gpp

markBowman:
As already supported by most other AOLs. For example libMyLibrary.so.2
Add compile scope to narUnpack mojo to correctly resolve dependencies.


ctrueden:
@markBowman Thank you very much! (And in case you were curious: the buildhive errors are an unrelated issue.)

markBowman:
Brilliant! Thanks. If it's not too cheeky, do you have any idea when you might push a new tag to maven central?

ctrueden:
@markBowman Since you asked, I tried to cut a `3.2.1` release just now.

Unfortunately, it failed. It seems [I don't have permission to deploy `com.github.maven-nar` artifacts to OSS Sonatype](http://jenkins.imagej.net/job/Release-Version/463/console). I [asked a Sonatype Nexus admin to add me](https://issues.sonatype.org/browse/OSSRH-7237) but in the mean time, perhaps @dscho could do the honors. Otherwise, I will cut the release once I have such permission.

markBowman:
Thank you. I really appreciate the excellent support.

ctrueden:
All sorted now, and [nar-maven-plugin 3.2.2](https://oss.sonatype.org/content/repositories/releases/com/github/maven-nar/nar-maven-plugin/3.2.2/) is released! Should appear on Central within a few minutes.

PR #151, opened 2015-02-24
Allow the use of the options -p and -pg for gcc linker.

dugilos:
Gcc linker allow the use of the options -p and -pg for debug purpose
(refer to the end of the message for documentation).

I configured those options in a pom:
<pre>
&lt;linker&gt;
	&lt;name&gt;g++&lt;/name&gt;
	&lt;options&gt;
		&lt;option&gt;-p&lt;/option&gt;
		&lt;option&gt;-pg&lt;/option&gt;
	&lt;/options&gt;
	...
&lt;/linker&gt;
</pre>
On linking step the options -p and -pg are preceded by "-Wl," on the
command line due to the fact that they are not registered in the
linkerOptions in the class com.github.maven_nar.cpptasks.gcc.GppLinker :
<pre>[DEBUG] g++ -Wl,-p -Wl,-pg ...</pre>

I add , "-p", "-pg" in linkerOptions in the classes GppLinker
and GccLinker to allow the use of those options.

-p Generate extra code to write profile information suitable for the
analysis program prof. You must use this option when compiling the
source files you want data about, and you must also use it when linking.
-pg Generate extra code to write profile information suitable for the
analysis program gprof. You must use this option when compiling the
source files you want data about, and you must also use it when linking.

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/gcc/GccLinker.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/gcc/GccLinker.java
@@ -42,7 +42,7 @@ public class GccLinker extends AbstractLdLinker {
     	    "-dynamic", "-arch",
             "-dynamiclib", "-nostartfiles", "-nostdlib", "-prebind", "-s",
             "-static", "-shared", "-symbolic", "-Xlinker",
-            "--export-all-symbols", "-static-libgcc",};
+            "--export-all-symbols", "-static-libgcc", "-p", "-pg"};
 // FREEHEP refactored dllLinker to soLinker
     private static final GccLinker soLinker = new GccLinker("gcc", objFiles,
             discardFiles, "lib", ".so", false, new GccLinker("gcc", objFiles,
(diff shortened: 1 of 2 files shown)


dugilos:
Hello, sorry I didn't present myself nor fill an issue form before creating this pull request, but this is a modification I did for myself and I feel it could be interesting to include it in the future release.

ctrueden:
@dugilos No problem at all! Sorry for the delay in review. The NAR project is quite undermaintained (too many questions/issues/PRs, too few maintainers, too little time).

ctrueden:
Oh hey, actually you just filed this. For some reason I was thinking it sat around for a long time and then your comment was a ping for attention. :grin:

ctrueden:
So there you go, super fast merge. Nice patch, great commit message, seems like an obvious addition (though I'm no expert on gcc).

dugilos:
Cool, that was fast !! :)
The message from buildhive fear me a little before I saw that the build is broken since january 5th on the clean install of it0027-dep-jni-native-lib-loader. I will take a look to see if I see anythink to correct that.

ctrueden:
Yeah, I haven't had time to look into the BuildHive situation since it0027. If you see an easy-yet-proper fix, that would be awesome.

PR #153, opened 2015-02-26
Allow the use of the option -pthread for gcc linker.

dugilos:
Gcc linker allow the use of the option -pthread for managing pthread
specificity when linking.

I configured this option in a pom:
<pre>
&lt;linker&gt;
	&lt;name&gt;g++&lt;/name&gt;
	&lt;options&gt;
		&lt;option&gt;-pthread&lt;/option&gt;
	&lt;/options&gt;
	...
&lt;/linker&gt;
</pre>
On linking step the option -pthread is preceded by "-Wl," on the
command line due to the fact that it is not registered in the
linkerOptions in the class com.github.maven_nar.cpptasks.gcc.GppLinker :
<pre>[DEBUG] g++ -Wl,-pthread ...</pre>

This commit add , "-pthread" in linkerOptions in the classes GppLinker
and GccLinker.

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/gcc/GccLinker.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/gcc/GccLinker.java
@@ -42,7 +42,8 @@ public class GccLinker extends AbstractLdLinker {
     	    "-dynamic", "-arch",
             "-dynamiclib", "-nostartfiles", "-nostdlib", "-prebind", "-s",
             "-static", "-shared", "-symbolic", "-Xlinker",
-            "--export-all-symbols", "-static-libgcc", "-p", "-pg"};
+            "--export-all-symbols", "-static-libgcc", "-p", "-pg",
+            "-pthread"};
 // FREEHEP refactored dllLinker to soLinker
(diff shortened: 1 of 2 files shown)


dugilos:
Hello, another link option that I added for me and that is worth to be shared I think.

ctrueden:
I noticed that the arm-Linux-gpp AOL [already adds `-pthread` by default](https://github.com/maven-nar/nar-maven-plugin/blob/9104c75d13a91d2b03c8df428d9877df4999443d/src/main/resources/com/github/maven_nar/aol.properties#L570). Should other AOLs also add this flag always? Or does it need to be configurable?

ctrueden:
(Either way: seems good to merge this. But it may be moot if it's always enabled for relevant AOLs.)

dugilos:
I didn't think of default conf from AOL, hopefully for me only arm-Linux-gpp defines -pthread so the fact to allow the use of the option stays correct (but it is unnecessary for amd64-Linux-gpp, on my computer at least, as gcc automatically links against pthread without any specific option).
Anyway, even if it's useless its not a problem to allow the use of this option I think, so I think we can keep it.

PR #160, opened 2015-04-16
Remove 'System.out.println' from NarSystem.java

jomarmar:
It is a bit annoying seeing a message in the console like: "your library cannot be found in java.library.path", when it is eventually loaded by NarSystem generated class (loaded by name, after extraction).


ctrueden:
Cool, thank you!

FWIW, I briefly looked at logging these messages in a less obnoxious way. The NAR plugin code itself has access to the Maven plugin logging framework (`org.apache.maven.plugin.logging.Log`), but `NarSystem` does not. We could use `java.util.Logger` there, but it feels like overkill. If someone complains later about these exceptions being silently eaten, we can add it then.

PR #163, opened 2015-04-25
Ccache

markjohndoyle:
Added support for compiling with ccache via a cpp property.

This change allows the nar maven plugin to compile using ccache:
(https://ccache.samba.org/)

It does this by adding a boolean property to the cpp config called ccache.
If this is true, the command is tweaked to call ccache with the compiler command.

Details:
 * Added boolean ccache maven parameter
 * Added ccache boolean flag to CompilerDef.
 * CommandLineCompiler tweaks the command if ccache in use.
 * CommandLineCompilerConfiguration passes useCcache to compiler compile method.
 * Existing unit tests use false for ccache.
 * Added some javadoc for getCompilerDef in Compiler.java
 * Updated configuration.apt with usage details.

There are also two commits containing a bit of clean up. Nothing major.

In the future this could possibly be converted to a compiler wrappers option to we can change things like distcc colourgcc too.

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineCompiler.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineCompiler.java
@@ -20,11 +20,11 @@
 package com.github.maven_nar.cpptasks.compiler;
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.List;
 import java.util.Vector;
-
-import org.apache.tools.ant.BuildException;
-import org.apache.tools.ant.types.Environment;
 
 import com.github.maven_nar.cpptasks.CCTask;
 import com.github.maven_nar.cpptasks.CUtil;
@@ -36,6 +36,10 @@ import com.github.maven_nar.cpptasks.TargetDef;
 import com.github.maven_nar.cpptasks.VersionInfo;
 import com.github.maven_nar.cpptasks.types.CommandLineArgument;
 import com.github.maven_nar.cpptasks.types.UndefineArgument;
+import com.google.common.collect.ObjectArrays;
+import org.apache.maven.surefire.shade.org.apache.commons.lang.ArrayUtils;
+import org.apache.tools.ant.BuildException;
+import org.apache.tools.ant.types.Environment;
 /**
  * An abstract Compiler implementation which uses an external program to
  * perform the compile.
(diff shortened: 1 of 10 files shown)


src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineCompiler.java
```diff
@@ -134,17 +140,24 @@ protected void buildDefineArguments(CompilerDef[] defs, Vector<String> args) {
     }
     /**
      * Compiles a source file.
+     * @param commandPrefixes 
      * 
      */
     public void compile(CCTask task, File outputDir, String[] sourceFiles,
             String[] args, String[] endArgs, boolean relentless,
-            CommandLineCompilerConfiguration config, ProgressMonitor monitor)
+            CommandLineCompilerConfiguration config, ProgressMonitor monitor, boolean ccache)
```
dscho: Both this and the `getCompiler()` rename are unnecessarily backwards-incompatible, and would therefore require a major version increase. The ccache only merits a minor version increase, though. Better leave redirects (for `getCompiler()` deprecated.

dscho:
This could use a test, and with the approppriate backwards-compatibility, the diff could be reduced dramatically. Also, the commits mix in whitespace changes which make reviewing and bisecting harder than necessary.

Otherwise fine (from a cursory read on a tablet).

src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineCompiler.java
markjohndoyle: No problem! What do you mean by redirects?

I assumed these were internal methods but I guess their public nature means you could have a dependent project?

markjohndoyle:
Oh, I thought I'd got all the whitespaces, sorry about that.

src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineCompiler.java
dscho: If they're `public`, they're `public` :smiley_cat:

markjohndoyle:
I removed the getCompiler - getComilerDef rename for now rather than deprecating getCompiler and redirecting to getCompilerDef.

I also removed the ccache boolean parameter from compile and exposed it in the config which is passed anyway.

I think that kind of work belongs in a separate commit - perhaps under an issue related to cleaning up the code, removing warnings and the like.


Hope that helps!

ctrueden:
This patch looks great. Unfortunately, I just merged a really disruptive style cleanup PR (#168). I'm trying to update this patch now; will reply back shortly with more details.

ctrueden:
OK, I cleaned up the patch series and filed a new PR for it: #169. Further discussion there.

Issue #167, opened 2015-05-05
Code formatter, organize imports and code cleanup


pauldaustin:
I'm not sure what IDE people use to develop the Nar plugin (I use eclipse).

On other projects I have an eclipse configuration for formatting and code cleanup. I then have eclipse to run the code cleanup before save. This means that methods are always ordered in alphabetical order, formatted according to the style guidelines and unused imports are removed or .* imports are converted to full imports.

By doing this it reduces the diffs on commits as it makes sure the formatting is consistent.

I also add a .gitattributes file to define the standard line endings (UNIX) for most files (*.java, *.xml) except windows specific files (e.g. *.bat, *.cmd) and binary (*.exe, *.png).

If this is something the team is interested in I can add sample config and instructions to a branch and create a pull request. Then you can review them.

FYI for the code formatting I basically follow the Sun Java Code Conventions http://www.oracle.com/technetwork/java/codeconvtoc-136057.html.

2 spaces for indent and no tabs. I've found 2 spaces is readable while compact compared with 4 spaces. Also placing the { on the same line for compactness.

dscho:
> On other projects I have an eclipse configuration for formatting and code cleanup.

NAR has that, too: https://github.com/maven-nar/nar-maven-plugin/tree/master/config

pauldaustin:
Great I have added that config to my eclipse to use for the project.

Could you add a Java\Code Style\Clean Up configuration too?

And a .gitattributes? See sample below

````
* text=auto

# text lf
*.c text eol=lf
*.css text eol=lf
*.cxx text eol=lf
*.h text eol=lf
*.i text eol=lf
*.ini text eol=crlf
*.java text eol=lf
*.jsp text eol=lf
*.properties text eol=lf
*.py text eol=lf
*.txt text eol=lf
*.xml text eol=lf
*.xsd text eol=lf
*.rglayer text eol=lf
*.rgobject text eol=lf

#text crlf
*.bat text eol=crlf
*.cmd text eol=crlf
*.ini text eol=crlf
*.exe.manifest text eol=crlf

#binary
*.exe binary
*.ftp binary
*.png binary
*.jpg binary
*.exe binary
*.ddl binary
*.exp binary
*.lib binary
*.jar binary
*.zip binary
*.dylib binary
*.pdb binary
````

ctrueden:
A PR adding that stuff, as well as maybe applying the styles across the codebase, would be very welcome!

[...]

PR #168, opened 2015-05-06
Eclipse code style and .gitattributes #167

pauldaustin:
Updated the eclipse formatting profile.

Added an eclipse code cleanup profile.

Added a screen shot of the eclipse save actions on a project.

Normalized line endings on files.

Ran the code cleanup and formatting on all the java source


ctrueden:
Thanks for this. I have cherry-picked the "big cleanup" commit (56e72df) onto master. This avoids the needless commit-then-revert series before it (0c2ee8e and ac5e466). There was also a problem on my OS X system with the "line endings" commits, so I redid that one as f261be803463315ad6249b96ed904a4c5fc74392.

dugilos:
Hello, I updated my local git reposity to get the plugin in its last version and I installed it (maven install) on my computer, Windows 7 64, and I have since some projects that doesn't compile anymore.

During compilations the library type of my projects is fixed to shared with no account of what I specified in the POM, I ran in debug mode to see what appends, and here is the result:
library-type
In the Library class instance, the attribute "type" seems to be correctly setted by the POM, the expression "this.type" seen by the debugger returns "executable", but the return of the method getType() is "shared" and this is the library type which is used during all the plugin execution.

The modification that brought this behavior is the "final" property of the "type" attribute.
before: <pre>    @Parameter
    private String type = SHARED;</pre>
now: <pre>  @Parameter
  private final String subSystem = "console";</pre>

By removing the property "final" my library type is back to normal during the plugin execution.
So I think that none of the @Parameter attributes should be final.

ctrueden:
Thanks @dugilos!

pauldaustin:
I'll look at this tomorrow. I'll remove the final flag from fields in the config and do a search in the code and manually remove the final keyword on parameters

Paul Austin
Revolution Systems Inc.

ctrueden:
@pauldaustin In case you didn't already notice: @dugilos already fixed it in PR #172. So no worries!

PR #169, opened 2015-05-14
Ccache

ctrueden:
Continuation of #163, which adds support for compiling with ccache via a cpp property.


ctrueden:
@markjohndoyle I took the liberty of splitting up your big commit into more atomic chunks. I also preserved all existing `CommandLineCompilerConfiguration` constructor signatures, to avoid breaking backwards compatibility with the existing API.

ctrueden:
Thank you very much for the contribution, @markjohndoyle!

markjohndoyle:
You are welcome!

PR #170, opened 2015-05-19
Msvc configuration

pauldaustin:
Added the new Msvc configuration class.

The configuration allows defining the MS Visual Studio version number.
This will then auto detect the Visual Studio Home. If the version is not
specified it will look for the VSxxxCOMNTOOLS environment variables to
see which versions exist and use the maximum available version.

The configuration also allows selecting the windowsSdkVersion. If not
specified it will look in the

C:\Program Files (x86)\Microsoft SDKs\Windows

or

C:\Program Files (x86)\Windows Kits

directories for available versions.

The linker and compilers will then be configured with default includes
and library paths from both the visual studio and windows sdk homes.

For example.

````
<msvc>
  <version>10.0</version>
  <windowsSdkVersion>7.1A</windowsSdkVersion>
</msvc>
````

### CODE REVIEW

**Comment by ctrueden** [comment]
There are still some problems. I'm cleaning up the patch now. I'll make a single commit in your name with your nice long explanatory commit message.

**Comment by ctrueden** [comment]
OK, I have a cleaned up patch, which I believe no longer has any extraneous changes. @pauldaustin Could you please take a look at the following patch and make certain it looks exactly how you want it to look?

b92c724771cf51915e1874115170dcac2f07cb27

In particular, please double check that all the hunks are actually relevant to this work. The patch is a bit monolithic and I'm not 100% certain that I didn't introduce any accidental reversions of unrelated work recently merged to master.

On [the topic branch](https://github.com/maven-nar/nar-maven-plugin/compare/msvc-configuration), I also added another commit that adds a `NarUtil.isWindows()` method, for convenience.

**Comment by pauldaustin** [comment]
Can you add back in the statements into Msvc that I had in the code? This makes it easier to read the code and verify that all my changes were included.

e.g.

if (home == null) {
  return false;
} else {
  :
}

instead of 

if (home == null) {
  return false;
}
  :

Sorry about the large commit. There wasn’t much way around that with this feature.

Paul

**Comment by ctrueden** [comment]
> Can you add back in the statements into Msvc that I had in the code?

Honestly I'm not worried about the `Msvc` class—that one is new, so is almost certainly the right code. I'm more concerned about the hunks touching existing files.

I don't have any more time to devote to this today. But of course you can "hard reset" your branch to match mine (`git fetch git://github.com/maven-nar/nar-maven-plugin msvc-configuration && git reset --hard FETCH_HEAD`) and then make additional changes on top, and/or squash them in. Then finally push back to this PR using `git push --force`.

**Comment by pauldaustin** [comment]
I’ve finished my diff against your msvc-configuration branch and everything looks fine.

**Comment by ctrueden** [comment]
Thanks @pauldaustin, I'll just merge the branch then!

**Comment by ctrueden** [comment]
Looks like these changes broke the Windows build:

http://jenkins.imagej.net/job/NAR-Maven-Plugin-windows/60/console

@pauldaustin Can you please investigate when you have time? I don't have a Windows system handy so it's more difficult for me to debug it.

**Comment by ctrueden** [comment]
The specific failure is:
```
com.github.maven_nar.test.TestLinkerVersion
testVersion(com.github.maven_nar.test.TestLinkerVersion)  Time elapsed: 0.062 sec  <<< ERROR!
org.apache.maven.plugin.MojoFailureException: Cannot deduce version number from: 
	at com.github.maven_nar.Linker.getVersion(Linker.java:491)
	at com.github.maven_nar.Linker.getVersion(Linker.java:425)
	at com.github.maven_nar.test.TestLinkerVersion.testVersion(TestLinkerVersion.java:56)
```

pauldaustin:
Add the code in red to line 57

final String version = this.linker.getVersion(new NarCompileMojo());

ctrueden:
@pauldaustin There is no "code in red" (if you tried to do color, I don't think GFM supports it).

The line `this.linker.getVersion()` still works because I left that public API method in place to avoid breaking backwards compatibility. And if you look at its [implementation in Linker](https://github.com/maven-nar/nar-maven-plugin/blob/56b3640d1e9a9918fcc1a1d43ee889797017d18f/src/main/java/com/github/maven_nar/Linker.java#L424-L426), it does nothing more than call `getVersion(new NarCompileMojo())`. So I would be surprised if your suggested change fixes the issue...

pauldaustin:
Do you know what version of Visual Studio is installed on the build server and in what directory?

ctrueden:
Microsoft Visual C++ 2010 Express at `C:\Program Files (x86)\Microsoft Visual Studio 10.0\Common7\IDE\VCExpress.exe`

`link /?` says `Microsoft (R) Incremental Linker Version 10.00.30319.01`

I'm happy to try running various things on the Windows build node... but not till after June 8, unless they are really quick.

pauldaustin:
Thanks,

I’ll look at this tomorrow.

PR #172, opened 2015-05-23
Remove the property final on the @Parameter attributes.

dugilos:
In a pom I defined a library type as static, and during the execution of
the plugin the library type was set to shared which is the default
value.
By removing the keyword "final" on the attribute "type" in the class
Library, the build was back to normal.
The attribute "type" has the annotation @Parameter and thus its value 
is normally set from the pom by maven, but the property final seemed to 
prevent that.

So I removed the keyword "final" on all the @Parameter attributes.

### DIFF
--- a/src/main/java/com/github/maven_nar/AbstractCompileMojo.java
+++ b/src/main/java/com/github/maven_nar/AbstractCompileMojo.java
@@ -87,7 +87,7 @@ public abstract class AbstractCompileMojo extends AbstractDependencyMojo {
    * Maximum number of Cores/CPU's to use. 0 means unlimited.
    */
   @Parameter
-  private final int maxCores = 0;
+  private int maxCores = 0;
 
   /**
    * Fail on compilation/linking error.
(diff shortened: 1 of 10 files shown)


dugilos:
Hello, after leaving a comment in the pull request #168, I tested the modification on my side and all the IT tests are back to success, so here is a pull request for this correction :)

ctrueden:
Thank you, makes total sense!

We may also want to change the Eclipse code style template to not insert final on private fields. We have this same problem in the ImageJ project, but it can be pretty easily sidestepped by avoiding the scenario where inserting final would make sense. E.g., if your private field does not assign a default value at declaration, and doesn't have exactly one assignment along each constructor code path, then it can't be marked final anyway.

ctrueden:
And yeah: if you mark a field final, it wreaks havoc with reflection. Even though there is a hack to remove the final keyword at runtime, after which you can assign the value (kinda sorta), it does not work as expected in many scenarios. So definitely best to avoid marking any `@Parameter` fields as final.

ctrueden:
I changed the style template: c0ad7e25f858ddcfd80c72705b379cae99022f19

PR #173, opened 2015-05-28
Ensure the Msvc class is initialized when retrieved from the

pauldaustin:
Ensure the Msvc class is initialized when retrieved from the
AbstractNarMojo. This is required when the Linker.getVersion() is used
with a dummy NarCompileMojo.

### DIFF
--- a/src/main/java/com/github/maven_nar/AbstractNarMojo.java
+++ b/src/main/java/com/github/maven_nar/AbstractNarMojo.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


ctrueden:
Thanks, I merged a cleaned up version of this patch to master: 4a34116f44bdd6782a28d7de5245f6d9ad745bd8

ctrueden:
Good news! It fixed the unit test failure. :+1:

Bad news! The ITs fail, with this message:
```
[ERROR] cl : Command line error D8037 : cannot create temporary il file; clean temp directory of old il files
```
See e.g. the [it0003-jni build log](http://jenkins.imagej.net/job/NAR-Maven-Plugin-windows/ws/target/it/it0003-jni/build.log).

Does the it0003-jni pass on your system? (See [this FAQ entry](https://github.com/maven-nar/nar-maven-plugin/wiki/Frequently_Asked_Questions#q-how-can-i-test-a-specific-integration-test) for how to run it.)

pauldaustin:
I think I may have at least partially figured it out.

I didn’t realize there is a getTestCompiler in addition to getCompiler.

To get the path setup correctly I have to use cmd /c to run a command rather than just running the command.

pauldaustin:
I now have the it0003-jni build running on my machine with an update to support the NarTestCompileMojo.

Let me know if it works on your server.

Also see this in case your TMP directory is full.

https://msdn.microsoft.com/en-us/library/aa983374%28v=vs.80%29.aspx?f=255&MSPPError=-2147217396

If it doesn’t work I’ll have a look at it on Monday.

ctrueden:
Thanks very much for your persistence, @pauldaustin!

However, I don't see a PR with your update. Could you please file it when you get a chance?

PR #176, opened 2015-06-05
Static libraries are now linked in TestCompile

Tubeliar:


ctrueden:
According to buildhive, this [doesn't compile](https://buildhive.cloudbees.com/job/maven-nar/job/nar-maven-plugin/337/com.github.maven-nar$nar-maven-plugin/console). Would be good to be certain it compiles for you with all unit tests [and integration tests passing](`mvn -Prun-its`).

Tubeliar:
My apologies. I have corrected the typo and the plugin now builds and tests ok on my system.

ctrueden:
@dugilos You changed this behavior in the other direction on April 4 (d316a0cc854cc6dd99985ef05095f3a00418fa29). What do you think? Should we revert it according to @Tubeliar's patch here?

Tubeliar:
Let me explain why i needed this.
I'm writing a program that uses several external libraries, but let's use OpenCV for this example. I want a completely independent executable, so all libraries are built static. I have written some functions that operate on OpenCV data structures and so OpenCV symbols appear in my header files and code.

The test code needs to call my functions and in order to be able to do that it would need to be linked to OpenCV itself. One of my functions returns a matrix whose values i wish to inspect in the test code. Without linking static libraries i don't think this is possible at the moment.

ctrueden:
Thanks for the explanation, @Tubeliar. Let's give @dugilos a few days to consider and respond. And I'll merge this patch after that, unless we reach a consensus otherwise. Sound good?

Tubeliar:
Sounds good :-)

dugilos:
Hello, I thought it would work, so I made some tests to understand what happened.

I created a static library static-lib-1, a static library static-lib-2 with a dependency on static-lib-1 and in the test of static-lib-2 I use a class from static-lib-1, and indeed I have a linking error.

To complete my test I created a shared library shared-lib-2 with a dependency on static-lib-1 and in the test of shared-lib-2 I use a class from static-lib-1, and it works.

So the dynamic libraries created by ld (I use GCC on Windows with MinGW) contain the content of the static dependencies, I counted on that, but the static libraries created by ar don't contain the content of the static dependencies, sorry I missed that.

I see two solutions, the simplest is to remove <code>&& !binding.equals(Library.STATIC)</code> as done by Tubeliar and as it was before, or a little more complex is to remove <code>&& !binding.equals(Library.STATIC)</code> only if the tested library is a static library. What do you think ?

Note : For the second solution I have to test my libraries static-lib-1 and shared-lib-2 with msvc compiler to see if it works as with GCC. I'll do that tomorrow evening, I'm in France and it is almost 01h AM now ^^

GregDomjan:
If your building a static lib that will use other static libs, the most common approach seems to be that they remain separate by default and all need to be linked into subsequent artifacts, an alternate approach is to make an 'uber' archive library.  
examples
http://stackoverflow.com/questions/2157629/linking-static-libraries-to-other-static-libraries
http://stackoverflow.com/questions/13492365/how-to-merge-two-windows-vc-static-library-into-one

It would be my preference that uber libs be a configured option and not the default.
As with building an uber jar there then needs consideration for handling resources
 - merge in headers?
 - marking/removing dependent libs that are merged in final pom.xml (like shade?)

We had some issue with this happening by default (using private modified plugin so can't confirm with current nar-maven-plugin) I believe related to maven_nar/NarCompileMojo.java  around line 287  all dependent libs are listed for linking - thus making an uber lib?


In testing a static lib, if a static lib is being built to 'test' then inclusion of the artifact (.lib/.a) won't show anything in that specific test,  only further use of that test artifact in subsequent build will be visible.

It would seem that TestCompile should follow the same rules and have the same available options for choosing.

dugilos:
@Tubeliar and @ctrueden I finished my tests with msvc, I have the same behavior than with GCC, but with the following drawback:
The class and/or functions in the static library wich will be embeded in the dynamic library must be marked by <code>__declspec(dllexport)</code> in order to be used directly from the dynamic library. It is rather logical but you don't need to do that with GCC as in Windows with MinGW all the symbols are exported by default from the dll (you can override that and export only some symbols), and on Unix systems there is nothing like <code>__declspec(dllexport)</code> for dynamic libraries.

And I just realised that I made a real mistake by not linking the static libraries to the test executable, you can't use anymore a static library as a test library (dependency with the scope test) for it will not be linked to the executable !

So I see 3 solutions:
1) Always link the static dependencies to the test executable.
2) Link the static dependencies to the test executable if the tested library is a static library (it will not contain the static dependencies), or if the dependency has the scope test.
3) Same as the point 2) above but with a new option in the &lt;test&gt; element in the POM to link the static dependencies anyway.

Note that with the point 2), has said above, with msvc, or GCC on Windows in certain conditions, the functions and / or classes in the static dependencies will need to have <code>__declspec(dllexport)</code> in order to be used through the dynamic library which will contain them.

It would be only for me I'd choose the solution 3) as I know how to handle it, but it is too complex I fear, what do you think ?

Tubeliar:
I need to clarify here, in my case i wasn't building a library. I am building an executable this is statically linked to its dependent libraries. For the test goal i want to perform unit tests by calling some functions that are included in the executable and validating the results. Therefore, the test application is an executable with dependencies on the static library.

So solution 1, always links, seems to be the correct one to me. Linking symbols that aren't used doesn't hurt, but not linking necessary libraries simply fails. Since a library is in de main dependency list, why would it not be a test dependency?

dugilos:
@Tubeliar You are right, I made the modification on 4 Apr to avoid linking to the test executable the static libraries that are already present otherwise. But it concerns only a specific case (the one I used obviously) and I didn't think further.
So I agree to your correction :) 

@ctrueden All is good for me then, you can merge this patch.

ctrueden:
@Tubeliar @dugilos Thanks for the discussion. I squashed the two commits together and pushed it to master.

ctrueden:
Also, it sounds like there are enough nuances to this issue that it might be worth adding a FAQ entry and/or wiki page about it. If anyone has the time, it would be much appreciated, and I'm sure help clarify matters in the future.

Issue #162, opened 2015-04-23
compiling c++ objects to a mirrored directory structure


Oneplus:
This issue is discussed in https://groups.google.com/forum/#!topic/maven-nar/SJOHdss-J-8

nar-maven-plugin compiles c/c++ files into same directory. When compiling files with same basename but under different directories, it gives a `Output filename conflict` error.

This makes it inconvenient when compiling c/c++ project with tons of same name files. Will it be possible that compiling files into mirrored directory structure to avoid such error, just like *scans* or *CMake* does.

ctrueden:
This would be a great feature. See also https://groups.google.com/d/msg/maven-nar/ptvPTa2vgkM/vtCCoC-468UJ

Oneplus:
@ctrueden Actually, we have our old project using `CMake` for the jni part and ant for the jar part. However, it's quite messy when we deliver the project. For some msvc user, they may have to use CMake to generate the `.vsproject` and compile it with visual studio via GUI.

So, my point is that CMake can be a quick and free solution. However, it cannot be a neat solution. Also it's expected that lots more work should be done than just fixing the NAR plugin as-is.

Oneplus:
@ctrueden see the same issue and a PR in #141 , but it seems fails the CI test.

ctrueden:
@Oneplus I think that PR #141 needs substantial work, unfortunately—beyond just commit cleanliness etc., the strategy the branch pursues to solve the problem looks to be too fragile. So I think further thought and work is needed to resolve this issue.

PR #178, opened 2015-06-11
A fix to "NAR: Compile failed: Output filename conflict"

Oneplus:
I generally follow #141 to insert a `hash(absoluteSourceFilePath)` between the object file basename and its extension. I found it quite difficult to mirror the folder structure of the source file into the object file (like discussed in #162 ), because the `compile` interface in `src/main/java/com/github/maven_nar/cpptasks/compiler/AbstractCompiler.java` make the output file invisiable to `commandlinecompiler`.

What #141 seemed to miss is that c++ compiler should compile one source into one object file by one command. But the original code seems to feed several sources to one command and it makes `-o` to several objects invalid.

It's a quick and dirty solution to the conflict problem however may works.


Oneplus:
Found it fails `it0018-fortran`. Could anyone advise on the possible reason?

ctrueden:
Presumably this is superceded by #179 now?

Oneplus:
@ctrueden Yes

PR #179, opened 2015-06-12
A fix to "NAR: Compile failed: Output filename conflict"

Oneplus:
Sorry for duplicate pr

This pr mainly involves inserting a `absoluteSourceFilePath.hashCode()` between the object file basename and its extension, trying to solve the "NAR: Compile failed: Output filename conflict" problem.

For example, an object fiie of `HelloWorldJNI.o` is transformed into `HelloWorldJNI.1824699243.o` (in `src/main/java/com/github/maven_nar/cpptasks/TargetMatcher.java`) to avoid output filename conflict. Such transformation requires an additional `-o` option (`/Fo` in MSVC) in commandlinecompiler which is introduced in `src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineCompiler.java`.

In previous nar-maven-plugin, several source files are compiled in one commandline. For example e `gcc a.c b.c c.c`. Since it relies on `gcc` to output `a.o b.o c.o`, to output `a.hashed.o b.hashed.o c.hashed.o`, I split it into 3 commands.

To make the compileOrder works (it0018-fortran), the comparison function on outputfilename extracts the basename (as `src/main/java/com/github/maven_nar/cpptasks/CCTask.java`)

Could anyone advise on this, since I am not sure about whether it is a good change.

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/CCTask.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/CCTask.java
@@ -37,6 +37,7 @@ import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.Task;
 import org.apache.tools.ant.types.Environment;
+import org.apache.commons.io.FilenameUtils;
 
 import com.github.maven_nar.cpptasks.compiler.CompilerConfiguration;
 import com.github.maven_nar.cpptasks.compiler.LinkType;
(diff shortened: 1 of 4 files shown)


ctrueden:
Thanks!!

I offered a couple of quick comments. But I don't really feel qualified to judge the core idea of this feature. Still, as long as it doesn't break existing use cases—and has a chance of avoiding output file name clashes in some circumstances—then I don't really see a problem with merging it even if the feature is imperfect or hacky.

But at the moment I think there may be bugs relating to the argument accumulation and counting, so I'd like to see that addressed first.

Oneplus:
@ctrueden Thanks for comments. 

Basically, this PR tried solve the conflict name problem with the smallest change on the existing project. So some feature is implemented in the same with the original project, like the `String[] commandline`.

For the suggestion on using List<String> instead of String[]
I've been thinking that this fix will be assigned to some core developer. If you suggest to make such change, I will work on that.

ctrueden:
Sorry for the super long delay on re-review. I still do not feel competent to judge whether this is really a good change, but since no one else is reviewing it... MERGE!

PR #181, opened 2015-06-17
Windows msvc

GregDomjan:
Some cleanup to recent addition of msvc config
Changes to support Windows specific compilations such as resource only dll with no manifest
Remove from aol config properties that where defined by compiler
Fix multi thread/core bug
Enhance to work in project with multiple dependency types.
Refactor artifact filtering and add option to config further filtering


ctrueden:
Cool, thanks for all the work. I have some concerns about breaking API changes. But in the interest of time, if you are confident these are all good changes, then I'm fine going to `4.0.0` with the next release.

ctrueden:
@pauldaustin Since a lot of this concerns your recent work, would you mind contributing to this PR with a code review? Personally I don't really feel comfortable judging the technical details.

GregDomjan:
I get where your coming from for semantic versioning and API, though I'm not feeling we have an API as such beyond the configuration. I'm not against this becoming 4.0.0.

ctrueden:
Going to `4.0.0` seems easiest to me. I don't care too much if we end up at 20.0.0, since as you point out there are not really any public users of NAR as a library. If that changes, we can start being more conservative about things like public method signature changes.

As for reviewing the functional meat of these changes: is there anyone besides @pauldaustin that we should invite to have a look?

GregDomjan:
Perhaps @dugilos could help review on changes for artifact selection/filtering.  

In the groups Thomas Haitzer recently asked about msvc and Henrik Horneber was asking about resource compilation.

ctrueden:
Might be worth posting to the group about this PR then, to invite other potential reviewers.

pauldaustin:
Look at the diffs for the following files. 

src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineCompiler.java
src/main/java/com/github/maven_nar/cpptasks/compiler/CommandLineLinker.java
You need code like the following otherwise the PATH doesn’t get passed to the command correctly. Ideally we should have a CommandLine executor wrapper that on windows would launch the command with cmd /c before the actual command.

if (NarUtil.isWindows()) {		
-      allArgsCount += 2;		
-    }		
    
-    if (NarUtil.isWindows()) {		
-      allArgs[index++] = "cmd";		
-      allArgs[index++] = "/c";		
-    }

Make sure the changes work with different windows SDK and MSVC versions. different windows SDK have different paths and include directories

GregDomjan:
Thanks @pauldaustin  
I have been testing mainly with VS 2013, but will check again with VS 2008, 2010, 2012. I don't seem to have 2009 installed anywhere... I've also been testing with SDK 7.0 through 8.1 

Including cmd needs to be unnecessary as there are too many issues with starting to use it, I would say it's far from ideal to introduce it. 
Command line introduces limits that need not be present when executing more directly, it also introduces possibility of injections and mangling/munging.
I'd like to look at using `org.codehaus.plexus:plexus-utils` CommandLine class as they have already dealt with many issues.  This might also be way forward with #179 

The PATH is now resolved correctly, and settings from the user are not accidentally injected causing or hiding problems.  Without the addition of the SystemRoot and TMP msvc would fail even with a good path.  Here is where using cmd was giving an injection already.  
Also required providing full path to the initial executable instead of relying on the cmd environment.

thohai:
Hi!
Unfortunately I am just a "user" of the msvc-code and did not even have a glance on it sofar.
So please take this into account when reading the next view lines:
I looked through the changes and could not find anything I would object to. I think that getting rid of the /cmd dependency is a good thing, especially if it means that the VC command prompt is no longer needed for executing the nar plugin. (This one I found rather unintuitive.)

On a side note: if I saw this correctly, this code is not ready for Windows 10 yet.
Hope this helps.

Thomas

pauldaustin:
@GregDomjan 

I assume by the following you mean that you set the tool path so that the command line has the full path to the command executed.

> Also required providing full path to the initial executable instead of relying on the cmd environment.

As long as the following still works I'm fine with the change

1. No need to launch the build from a Visual Studio command prompt with support for the auto-detection of the Visual Studio location.
2. Support for both x86 and 64 bit systems
3. Support for cross compiling (so x84 from a 64 bit compiler)
4. Support for SDK versions 7.1A, 8.0, 8.1
5. Support for at least VS 2008, 2010, 2013

GregDomjan:
Previously I missed an issue with mt.exe not using toolpath when changed my build to a visual studio cmd prompt. 

There are some differences between Windows SDK 6.0A/7.0A/7.1A (installed with visual studio) and 6.0/7.0/7.1 (stand alone windows platform sdk) but I'm not sure if it's acceptable to configure say 7.1A and 'automatically upgrade' to 7.1 - probably not good to 'downgrade' from 7.1 to 7.1A.  To me they are generally interchangeable so it feels like it should be more flexible instead of pedantic.
Since 8 visual studio seems to just install the full platform kit and so variation is no longer present.

GregDomjan:
@thohai I'm not aware of any issue this may face with running on Windows 10, perhaps you could elaborate?

ctrueden:
@GregDomjan @pauldaustin @dugilos So the consensus so far is that all of these changes are good as-is?

If so: @GregDomjan: if you have time to fix the merge conflicts, we can go ahead and merge whenever you're ready.

dugilos:
Hello, the IT test 28 that I made to test the changes I made is still OK, so for me there is no reserve.

ctrueden:
@GregDomjan The merge conflicts were super easy to resolve, so I just went ahead and did it. :smile:

ctrueden:
Continuing the shotgun approach of merging all the PRs, and then fixing any later brokenness if/when it arises. Especially since this one has a lot of nice cleanup. Thanks @GregDomjan!

GregDomjan:
Thanks, was just getting around to doing the same.

PR #183, opened 2015-07-22
System.mapLibraryName() changed in JDK 7 on OS X

diyessi:
The generated NarSystem.loadLibrary() uses System.mapLibraryName() to
search for the JNI library.  In JDK 7 on OS X, System.mapLibraryName()
switched from a file extension of .jnilib to .dylib so OS X linking needs
the corresponding change.

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/gcc/GccLinker.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/gcc/GccLinker.java
@@ -55,7 +55,7 @@ public class GccLinker extends AbstractLdLinker {
       false, null);
   private static final GccLinker machDllLinker = new GccLinker("gcc", objFiles, discardFiles, "lib", ".dylib", false,
       null);
-  private static final GccLinker machJNILinker = new GccLinker("gcc", objFiles, discardFiles, "lib", ".jnilib", false,
+  private static final GccLinker machJNILinker = new GccLinker("gcc", objFiles, discardFiles, "lib", ".dylib", false,
       null);
   // FREEHEP added dllLinker for windows
   private static final GccLinker dllLinker = new GccLinker("gcc", objFiles, discardFiles, "", ".dll", false, null);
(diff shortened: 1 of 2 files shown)


ctrueden:
This represents a conscious choice to (partially) abandon support for Java 6, at least on OS X. I am personally fine with that, but it might we worth a mail to the NAR mailing list first. Obviously, Java 6 is long obsolete, but at the moment NAR still supports it.

ctrueden:
An alternative might be to check the version of Java, and set the extension accordingly in the code.

diyessi:
The current situation is that on OS X the plugin does not work when used with JDK 7 or later.  Judging from what I found on the web, it doesn't seem like the plugin has been used on OS X for some time, hence the simple change.  A more complicated change would be to modify the generated code to check for both .jnilib and .dylib when mapLibraryName() returns .dylib, but that seems like overkill and somewhat riskier, but I can send mail to the NAR list if you think there is a chance it would be a problem.

ctrueden:
@diyessi OK, to keep things moving, I proposed an alternative approach. If you like it, feel free to amend the PR. (You can do this by using `git commit --amend` followed by `git push --force` on your topic branch locally.)

diyessi:
Making it depend on the vendor of the jvm doing the build feels like asking for trouble if there is more than one user for the nar.  It would be better to add a minimum java version configuration parameter that controlled .jnilib/.dylib and make the generated library loading helper code search for .jnilib if .dylib wasn't found. Or just don't support JDK6 on OS X.  I don't really have a way to test that jvm.

diyessi:
Another reason not to do it your way is that jni is supposed to be a vendor-neutral, so the change from .jnilib to .dylib would be associated with the version, not the vendor.  Java's >1.6 loader will look for .jnilib if it doesn't find .dylib.  I suppose there must be a way for the nar plugin to look in the java compiler plugin and see what the minimum java version is set to so that the linker could generate the extension according to that.

diyessi:
I tried changing the generated  NarSystem class to check for .dylib/.jnilib, which would potentially allow support for JDK6 and earlier, except that the native-lib-loader only handles the <7 case where mapLibraryName returns .jnilib but the library is .dylib.  In 7 or later, mapLibraryName returns .dylib so native-lib-loader won't find a .jnilib, so native-lib-loader would need a corresponding fix to make the make such a patch worthwhile.

ctrueden:
So, now that #184 and scijava/native-lib-loader#8 are both merged, this PR is obsolete and can be closed, right?

diyessi:
I can double-check on the merge with my test case in the next day or so 
if you want to be extra safe, but I think we're good.

ctrueden:
@diyessi Great, thanks. Closing; please reopen as needed.

PR #184, opened 2015-07-24
Check for .jnilib and .dylib on OS X

diyessi:
On OS X, through JDK 6 JNI libraries used .jnilib, but have since switched
to .dylib.  Change the generated NarSystem.loadLibrary that works with
native-lib-loader to work with either extension.

Note: This depends on a recent change to native-lib-loader to make it work with both extensions.


ctrueden:
I am utterly swamped through mid-September, with no time to do any testing. But in the interest of this change seeing the light of day, I'm going to "merge first, ask questions later." If it causes trouble for anyone, I'm sure we'll hear about it eventually. :wink:

PR #185, opened 2015-08-04
Do not check non-jar or non-nar artefacts for NarInfo

markBowman:
During the nar-download phase the plugin hunts for Nar dependencies. It does this by opening each dependency artefact using the jar application and looking for NarInfo. If any dependencies are unreadable by jar a ZIP exception is generated. This pull request adds as explicit check that each artefact is of type jar or nar prior to passing to the jar application. If the check fails a log message is generated and getNarInfo() returns null, forcing  getNarArtifacts() to move onto the next artefact.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,6 +244,11 @@
       <artifactId>xml-apis</artifactId>
       <version>1.3.04</version>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.4</version>
+    </dependency>
 
     <dependency>
       <groupId>junit</groupId>
--- a/src/main/java/com/github/maven_nar/AbstractDependencyMojo.java
+++ b/src/main/java/com/github/maven_nar/AbstractDependencyMojo.java
@@ -247,6 +247,11 @@ public abstract class AbstractDependencyMojo extends AbstractNarMojo {
       return null;
     }
 
+    if (!dependency.getType().equals("jar") && !dependency.getType().equals("nar")) {
+       getLog().debug("Skipping unreadable artifact: " + file);
+       return null;
+    }
+
     JarFile jar = null;
     try {
       jar = new JarFile(file);
--- a/src/main/java/com/github/maven_nar/AbstractDependencyMojo.java
+++ b/src/main/java/com/github/maven_nar/AbstractDependencyMojo.java
@@ -20,11 +20,16 @@
 package com.github.maven_nar;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.jar.JarFile;
+import java.util.zip.ZipException;
+import java.util.zip.ZipInputStream;
+import org.apache.commons.io.IOUtils;
 
 import org.apache.maven.artifact.Artifact;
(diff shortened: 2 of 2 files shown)


ctrueden:
I like this idea. But is there a more robust way than file extension to do this? What about other variants like `war`, `ear`, etc.?

Would it be sufficient to do:
```java
boolean isZipped = new ZipInputStream(yourInputStream).getNextEntry() != null;
```
and then perhaps also catch `IOException` here, and in later code, to move on if something goes wrong in the processing?

See also [SO #22401183](http://stackoverflow.com/a/22401183/1207769).

markBowman:
Something more like this?
```
ZipInputStream zipStream = null;
try {
        zipStream = new ZipInputStream(new FileInputStream(file));
        if (zipStream.getNextEntry() == null) {
            getLog().debug("Skipping unreadable artifact: " + file);
            return null;
        }
}
catch (IOException e) {
      throw new MojoExecutionException("Error while testing for zip file " + file, e);
}
finally {
        IOUtils.closeQuietly(zipStream);
}
```

GregDomjan:
I had something similar in the msvc pull request, rather than opening up to looking in any jar/zip I was tightening down to only nar dependencies.

https://github.com/maven-nar/nar-maven-plugin/commit/44386829391e1792bb85762282a01c21efa672c9

ctrueden:
@markBowman That approach looks good to me!

Always best to have a unit test when possible, but I don't think it is crucial here.

FYI, my schedule through the end of the summer is totally insane, and it'll be hard for me to seriously review any PRs before October. Perhaps other NAR devs (@GregDomjan @dproteus13 @richardkerr ?) can review and merge things in the meantime?

@GregDomjan I already offered my comments on PR #181; you can do what you want. E.g., if you want to merge it as-is and make the next release be `4.0.0`, that is OK with me, since I do not have the time to fight for total backwards API compatibility here.

richardkerr:
@ctrueden I've somewhat recently moved jobs which means no active projects using the plugin and very little time to dedicated to it, as is evident by my activity over the past year or more, and realistically this isn't going to change in the foreseeable future.  So that said, there would be no hard feelings if it was felt appropriate to remove my status as an owner of the repository.

ctrueden:
Thanks for the update, @richardkerr. If you don't mind I'll leave you on the project, just for the increased bus factor since you know the code.

markBowman:
I've modified my pull request to comply with @ctrueden comments, but in light of @GregDomjan patch in https://github.com/maven-nar/nar-maven-plugin/commit/44386829391e1792bb85762282a01c21efa672c9 I'm not sure which is the preferred solution.

I'd like to say again how much I appreciate all of your efforts to maintain this plugin. We use it extensively (http://lcogt.net) and working with it now is a lot easier than during the wilderness years.

richardkerr:
Fine by me @ctrueden

dproteus13:
Looks like Build Hive is still pending, I'll check again tonight: https://buildhive.cloudbees.com/job/maven-nar/job/nar-maven-plugin/

dproteus13:
Still pending, anyone know if it's just hosed?   I'll pull it down on my system for a manual test shortly just in case, I have a Windows tablet to test on. .. anyone else have a particular platform they want tested?

dproteus13:
Sorry I haven't been able to test yet, I'm running in to some issues with the configuration on my tablet...

ctrueden:
In the interest of time—and since this is still a fairly simple patch—I'm merging as-is.

@dproteus13 and/or anyone: Testing would still be appreciated if you have the time.

PR #188, opened 2015-09-16
Rework after last set of merges, Additional MSVC features

GregDomjan:
Again removing usage of cmd.exe on windows
Re factoring of the Conflicting output names, original was causing windows resource and idl compiles to fail.

Include MSVC symbols for static builds into the build artifact
Support for Windows Platform SDK 10 with multiple installed versions 'evergreen'


ctrueden:
Thanks @GregDomjan and sorry for the delay. September was an extremely busy month.

PR #189, opened 2015-10-13
Native lib loader searching also the "shared" directory and not only "jni"

thohai:
Hi!

I have a project setup with 3 projects where I use JNI to access a shared library using native lib loader.
However, the nativelibloader does not find the shared library as it only searches the JNI folder but not the shared folder thus resulting in an error that the library could not be loaded because the dependency was not found.

For me, this small change solves the problem.

Best
Thomas

### DIFF
--- a/src/main/java/com/github/maven_nar/NarSystemMojo.java
+++ b/src/main/java/com/github/maven_nar/NarSystemMojo.java
@@ -102,6 +102,9 @@ public class NarSystemMojo extends AbstractNarMojo {
     builder
         .append("            final File file = new File(prefix + aol + \"-jni/lib/\" + aol + \"/jni/\" + mapped);\n");
     builder.append("            if (file.isFile()) return file;\n");
+    builder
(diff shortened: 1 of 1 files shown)


ctrueden:
Thanks @thohai. The first commit looks very good, but the second one mixes in tons of changes which I think you didn't intend as part of that patch. And actually, it appears to be a completely redundant patch with the current master.

Anyway, I merged a slightly cleaned up version of your patch to master. So hopefully this issue is resolved for you now.

thohai:
Hi!

Thank you for cleaning up!
I intended only the one change in the NarSystemMojo anything else was 
not intended.

Best
Thomas

thohai:
Hi! What jumped my mind: what about static depenencies in this context .. would there be a problem too?

ctrueden:
@thohai I am far from an expert on native code. But I thought static linkage meant that dependencies are included in the resultant main binary? So they wouldn't be stored separately in the NAR, then?

thohai:
@ctrueden ofc you are right ... complete blackout on my side .....

Issue #190, opened 2015-10-25
Fails on latests master branch - looks for gcc in Visual studio directories.


markjohndoyle:
Hi all,

I've built the latest snapshot master branch and get the following error when compiling a project that uses gcc. I understand it's the snapshot but I thought it is worth tracking.

OS: Windows 10 x64
λ gcc --version
gcc.exe (Rev5, Built by MSYS2 project) 4.9.2

ERROR:
------------------------------------------------------------------------
Failed to execute goal com.github.maven-nar:nar-maven-plugin:3.2.4-SNAPSHOT:nar-compile (default-nar-compile) on project libdispatch: NAR: Compile failed: Could not launch C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\gcc: java.io.IOException: Cannot run program "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\gcc" (in directory "C:\Users\markj\code\netcdf-c-nar\libdispatch\target\nar\obj\amd64-Windows-gcc"): CreateProcess error=2, The system cannot find the file specified -> [Help 1]

markjohndoyle:
Looks like it's related to this new piece of code in Compiler.java

// tool path
    if (this.toolPath != null) {
      compilerDef.setToolPath(this.toolPath);
    } else {
      mojo.getMsvc().setToolPath(compilerDef,getLanguage());
    }

markjohndoyle:
I'll see if your commit helps, Greg.  I did a quick hack to check what happens if we bypass setting that MSVC toolpath myself and the CreateProcess still failed; it was a quick check though.

markjohndoyle:
@GregDomjan 
Yeah, unfortunately, it still fails, I've extracted the command lines used from the current build and a build from before the msvc branch merge.

## Compiler command lines
### Latest master branch build
[...]

PR #192, opened 2015-10-26
Issue #190

markjohndoyle:
Non-compliant environment configuration was forced on GCC compiler.

 * Removed line that forced MSVC environment configuration.

### DIFF
--- a/src/main/java/com/github/maven_nar/NarCompileMojo.java
+++ b/src/main/java/com/github/maven_nar/NarCompileMojo.java
@@ -215,8 +215,6 @@ public class NarCompileMojo extends AbstractCompileMojo {
     // add java include paths
     getJava().addIncludePaths(task, type);
 
-    getMsvc().configureCCTask(task);
-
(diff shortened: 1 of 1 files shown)


GregDomjan:
As you noted, simply removing this configuration would break msvc builds.
With this additional input have limited addition of msvc environment to when using the msvc linker.

markjohndoyle:
But leaving it in breaks GCC builds. It feels odd that there is MSVC specific code in the NarCompileMojo but I assume there is a good reason for this.

It's worth noting that the current code doesn't just add Visual Studio specific entries to the PATH, it replaces everything. I think this will break all compilers. Maybe the configure CCTask logic can be fixed to append the entries.

This code doesn't seem to affect the linker, I actually have another local [patch](https://dl.dropboxusercontent.com/u/10007025/linker-fix.patch) to get the linker working.

markjohndoyle:
I see you've made some follow up commits; i'll check them out. Thanks, Greg.

PR #193, opened 2015-10-27
Issue #190 - MSVC settings breaking other compilers

markjohndoyle:
* Check for MSVC linker before changing tool (linker) path.
 * Check for MSVC linker type before changing linker task.


markjohndoyle:
Linker version of the previous compiler changes, @GregDomjan 

This assumes the linker name for msvc is also "msvc".

if (os.equals(OS.WINDOWS) && "msvc".equalsIgnoreCase(mojo.getLinker().getName()))

ctrueden:
@GregDomjan Thank you for being on top of this!

PR #196, opened 2015-11-02
Switch from BCEL 5.2 to BCEL 6 to support Java 8 classes

stephanmehlhase:
Bug [BCEL-173](https://issues.apache.org/jira/browse/BCEL-173) causes
the nar-javah step to fail when used on classes using the
`invokedynamic` statement. Switching to BCEL 6 solves this problem.


ctrueden:
Great, thank you!

PR #197, opened 2015-11-04
Prefix for gcc g++ based compiler

gwilga:
This change add:

 - new tag ```<prefix>``` into ```<linker>```, ```<cpp>```, ```<c>```, ```<fortran>``` section.
Now we can compile and link with a cross-compiler or a custom compiler.

 - change behavior for the fallback compiler in CCTask, if there is no compiler define then use the fallback one. I wonder if this change should not be into another commit, but the feature is dependent on this modification.

Signed-off-by: Grégory Wilga <gwilga@gmail.com>

### DIFF
--- a/src/main/java/com/github/maven_nar/Compiler.java
+++ b/src/main/java/com/github/maven_nar/Compiler.java
@@ -63,6 +63,12 @@ public abstract class Compiler {
   private String name;
 
   /**
+   * The prefix for the compiler.
+   */
+  @Parameter
+  private String prefix;
+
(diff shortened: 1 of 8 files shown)


axel3rd:
+1 for this feature, very useful ! :+1:

ctrueden:
Thanks very much, @gwilga, and thanks for the nudge, @axel3rd. The patch looks good.

@gwilga It would be great if you could add a test in `src/it` to make sure this feature never breaks in the future.

PR #198, opened 2015-12-01
Add more path to look for installed windows SDK.

gwilga:
Regarding official documentation for windows SDK, installation can be into "C:\Program Files".
https://msdn.microsoft.com/en-us/library/ms717422%28v=vs.110%29.aspx
https://msdn.microsoft.com/en-us/library/ms717422%28v=vs.100%29.aspx

We should be able to check this directory too.

Signed-off-by: Grégory Wilga <gwilga@gmail.com>


GregDomjan:
Might be nice to toggle this based on OS platform 32/64 bit rather than just trying for both.
Will merge as is if no further issue is raised in next day.

PR #200, opened 2016-01-03
Change the algorithm to select the Windows SDK

kristofmoens:
In my current windows SDK installation, I only have windows 10 ucrt folder. I needed to change the algorithm to be able to compile with this SDK.

Will now search for the correct library paths, by combining the windows 10 SDK's with the previous
SDK's. This is done by selecting Windows 10 SDK's above the older ones; i.e. search first in Windows 10 library (newest first) and the next libs will be 8.1, 8.0, 7.1. This should make the algorithm more general.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
   <groupId>com.github.maven-nar</groupId>
   <artifactId>nar-maven-plugin</artifactId>
-  <version>3.3.0-SNAPSHOT</version>
+  <version>3.3.0</version>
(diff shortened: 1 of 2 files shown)


ctrueden:
Thanks @kristofmoens! I am going to clean up this patch to separate out unrelated changes (whitespace etc.) into their own commits. I'll reply back here and close when it's on master.

ctrueden:
Closed in favor of #201.

PR #174, opened 2015-05-31
Msvc init

pauldaustin:
Added the Msvc CCTask configuration to the NarTestCompileMojo and NarVcprojMojo. This should allow the tests to find the compiler using the path and complete.

PR #201, opened 2016-01-03
Add support for Windows 10 SDK

ctrueden:
This is a continuation of #200, contributed by Kristof Moens.

### DIFF
--- a/src/main/java/com/github/maven_nar/Msvc.java
+++ b/src/main/java/com/github/maven_nar/Msvc.java
@@ -280,8 +280,7 @@ public class Msvc {
           }
         }
         // TODO: else Registry might be more reliable but adds dependency to be
-        // able to acccess - HKLM\SOFTWARE\Microsoft\Visual
-        // Studio\Major.Minor:InstallDir
+        // able to acccess - HKLM\SOFTWARE\Microsoft\Visual Studio\Major.Minor:InstallDir
       }
       mojo.getLog()
           .debug(String.format(" VisualStudio %1s (%2s) found %3s ", this.version, internalVersion, this.home));
--- a/src/main/java/com/github/maven_nar/Msvc.java
+++ b/src/main/java/com/github/maven_nar/Msvc.java
@@ -2,13 +2,17 @@ package com.github.maven_nar;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.ListIterator;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.Comparator;
 
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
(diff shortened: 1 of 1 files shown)


ctrueden:
I cleaned up the patch to remove irrelevant style and whitespace changes. It is now minimal enough to enable review, I think.

@GregDomjan As our resident Windows guru, it would be awesome if you could have a look and/or test this on your machine(s). And actually, that goes for anyone here using Windows—I am not a Windows developer personally, so testing is more difficult for me.

In general, I do wonder how we can continue to maintain backwards compatibility successfully for all these old Windows MSVS versions. Perhaps we should install all the old versions of MSVS on our Jenkins Windows VM, and run the ITs against all those versions? Unfortunately, the ITs are failing on Windows for quite some time now (see #174 for discussion), so that would need to be addressed as part of this...

ctrueden:
Also, I forgot to say: nice cleanup, @kristofmoens! There are many great things about this patch. I think it is a step forward, organizationally.

kristofmoens:
Hi @ctrueden, about the `sdkInclude`: I have no idea actually if the order of the includes would matter. I only tested this with a simple project and I don't know of any overlap of certain include files in the Windows SDK, that why I just picked a set. A list could be a more efficient way of storing the list, so that's ok for me to change. :-) Thanks for the feedback!

kristofmoens:
So maybe adding a function that will skip the versions until you're at the current version? 
For instance: windowsSdkVersion:8.0 
You have a (sorted) list of all the SDK's installed, so you can iterate through them until you have arrive at versions <= windowsSdkVersion. This would give you the following process for the example:
10 -> skip
8.1 -> skip
8.0 -> continue as normal

kristofmoens:
Oh I see, I don't have extra folders in the 8.1 folder, I just choose to pick the first one since that was working in my (simple) case. It's rather difficult to get an overview of all possible folder combinations. Maybe the best would be to look for the lib folder (and architecture)? I won't have time to implement it at the moment since I have some other priorities now.

Also maybe it would be better to generate MSBuild files and use this to build it since this tool should hopefully know where to find everything? Wouldn't this be better than to try to implement all the (searching) logic ourself?  It would also allow for more intelligent multi-threading support, since now (if I understand correctly) the Windows SDK's are recalculated every time.

A good tutorial would be this:
https://msdn.microsoft.com/en-us/library/dd293607.aspx
I think I may prefer this solution since it would be less code to maintain and may give some performance improvements.

Maybe some of it could be reused to generate CMake build files?

ctrueden:
> Also maybe it would be better to generate MSBuild files and use this to build it since this tool should hopefully know where to find everything? Wouldn't this be better than to try to implement all the (searching) logic ourself?

For sure—personally I think NAR tries to do way too much. Better to lean on an existing tool.

> Maybe some of it could be reused to generate CMake build files?

Yes, CMake would be much better than generating MSBuild directly. If NAR generated `CMakeLists.txt` and then invoked `cmake`, that would be really slick. But that would be massive change, of course.

ctrueden:
I would love to merge this, but it is still waiting on feedback from someone else who uses NAR with Windows...

@GregDomjan any chance you could quickly test it out on one of your systems?

GregDomjan:
I hope to have a look soon. 
Just tried 3.3.0 and found it has some issue with the input/output location changing breaking some of the IDL compilation, so I'll try and fix that while verifying this.

ctrueden:
Thank you @GregDomjan !

PR #205, opened 2016-02-07
Change compiler working directory

### LINKED ISSUE #158: Full path in log
Hi,
We have issue with logs when we use this plugin.
In logs of running application we see full path to file on machine where was made compilation. File name is displayed using _FILE_ c defined macro

for example
if we compile the code by giving complete path to gcc
>  gcc -c /u02/dir/dir2/dir4/sources/foo.cc

this will show complete path to file in logs
>  "/u02/dir/dir2/dir4/sources/foo.cc"

but if we do "cd /u02/dir/dir2/dir4/sources/" and  say "gcc -c foo.cc"
in logs we see only file name.

How can we execute "gcc" without full path to source file? Maybe we can set dir for sources?

And the real problem is that if we take this compiled binary and execute it on other server, it will shows the same full path (from evn where we've compiled it) for us in logs for example.
Is it possible to add any option like workDir in plugin configuration?

dmitrykirsanov:
At current moment compiler command is executing in object directory,
where results of compilation (object files) are placed
(e.g. ${project.build.directory}/target/nar/obj/amd64-Linux-gpp).

It is more logical to use sourceDirectory as compiler working directory.
This allows cut path to source files and makes compiler command some shorter.

This PR can resolve #158 issue.

### DIFF
--- a/src/main/java/com/github/maven_nar/Compiler.java
+++ b/src/main/java/com/github/maven_nar/Compiler.java
@@ -464,7 +464,16 @@ public abstract class Compiler {
         compilerDef.addFileset(fileSet);
       }
     }
-
+    
+    if (type.equals(TEST)) {
+      if (this.testSourceDirectory.exists()) {
+        compilerDef.setWorkDir(this.testSourceDirectory);
+      }
+    } else {
+      if (this.sourceDirectory.exists()) {
+        compilerDef.setWorkDir(this.sourceDirectory);
+      }
+    }
     return compilerDef;
   }
 
--- a/src/main/java/com/github/maven_nar/cpptasks/CCTask.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/CCTask.java
@@ -46,6 +46,8 @@ import com.github.maven_nar.cpptasks.compiler.Linker;
 import com.github.maven_nar.cpptasks.compiler.LinkerConfiguration;
 import com.github.maven_nar.cpptasks.compiler.Processor;
 import com.github.maven_nar.cpptasks.compiler.ProcessorConfiguration;
+import com.github.maven_nar.cpptasks.compiler.AbstractCompiler;
+import com.github.maven_nar.cpptasks.compiler.CommandLineCompilerConfiguration;
 import com.github.maven_nar.cpptasks.ide.ProjectDef;
 import com.github.maven_nar.cpptasks.types.CompilerArgument;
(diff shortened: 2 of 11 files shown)


ctrueden:
Beautiful patch. Without having tested it, it looks good to me. Do all the integration tests still pass for you?

One concern raised by @GregDomjan during the discussion: what if there are multiple source directories, due to e.g. `build-helper-maven-plugin` configuration?

     /base/project/src/main/cpp/somefile.cpp
     /base/project/target/generated/cpp/somefile.cpp

This sort of thing occurs when parts of the source are autogenerated (e.g., by templates).

Do you think this patch would cause any problems in that scenario? Does anyone in the NAR community have a project which does this, which they could test?

dmitrykirsanov:
I have tested it on Unix with gcc compiler, on Windows with Visual Studio and Cygwin. It worked everywhere. All the integration tests was passed for me also.
I created a small test for checking multiple source files compilation in different directories (if necessary, I can share it with you). To make it work it is enough to set
`<sourceDirectory>${project.basedir}</sourceDirectory>`

ctrueden:
Great, thank you!

manu4linux:
@dmitrykirsanov @ctrueden 

Hi Dmitry, I know it's been a while you did this changes but we see that all the debug string filenames with in executable now only have relative path to main. for example `/base/project/src/main/c/somefile.c` would show as `c/somefile.c`

since our project use a static dependency libs, let say file from `/base/project2/src/main/c/someotherfile.c` will resolve to `c/someotherfile.c` causing GDB debugging issues. if I am not wrong Behaviour in old `maven-nar-plugin`seem to give full path which made it easy for GDB to find them. 

Hacking `GccCompatibleCCompiler.java` with below code seem to work for us..
 ```cpp
 try {
          relative = FileUtils.getRelativePath(workDir, new File(filename));
      } catch (Exception ex) {
     }
     if (relative.isEmpty()) {
         return filename;
     } else {
 +         return filename;  //  return relative;
```
based on your commit e816d1c i feel this has to be done on few other files too. Am i in right direction.

So this leave me asking what other advantages do we have in small relative path apart from size. 

Does having a switch to enable `relative path` or `full path` or even having some kind of `minimal path` from main would make sense.

manu4linux:
FYI i have also posted this at https://groups.google.com/forum/#!topic/maven-nar/sIBkEOdjDfA

PR #206, opened 2016-02-14
Fix JNI library extensions on OSX

kdubb:
Since Java 7 OSX JNI libraries use the extension "dylib" instead of "jnilib".  Considering Java 6 is almost completely EOL, just switching the extensions without any specialized detection of Java environment should be sufficient.


kdubb:
This was not the correct fix and it seems somebody has shimmed a fix for loading into the generated NarSystem between 3.2.3 & the current 3.3.0-SNAPSHOT.

ctrueden:
FYI, I believe the changes you are talking about happened in #184.

PR #207, opened 2016-02-17
Add AIX OS and LIBPATH environment variable during unit test

picpicpic:
Hi,

Here is a pull request in order to set LIBPATH instead of LD_LIBRARY_PATH when running tests on AIX OS.

Any question or comments, please, don't hesitate.
Regards
Aldric


ctrueden:
Thanks!

PR #209, opened 2016-02-18
Make Compiler implement the toString() method

### LINKED ISSUE #204: com.github.maven_nar.IncludePath should override toString.
In `Compiler`, if an include path doesn't exist on the system, the code throws an exception and puts the not-found `IncludePath` into the message ... but because `IncludePath` does not override `toString` to produce any meaningful output, the user only sees something like `com.github.maven_nar.IncludePath@6f94fb9d` (the default output from `Object.toString`) and that makes it hard to know what include path wasn't found.

ctrueden:
@jcflack This is 100% untested; could you please give it a try?

### DIFF
--- a/src/main/java/com/github/maven_nar/Compiler.java
+++ b/src/main/java/com/github/maven_nar/Compiler.java
@@ -21,6 +21,7 @@ package com.github.maven_nar;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -618,5 +619,10 @@ public abstract class Compiler {
 
   public final void setAbstractCompileMojo(final AbstractCompileMojo mojo) {
     this.mojo = mojo;
+  }
+
+  @Override
+  public String toString() {
+    return NarUtil.prettyMavenString(this);
   }
 }
--- a/src/main/java/com/github/maven_nar/NarUtil.java
+++ b/src/main/java/com/github/maven_nar/NarUtil.java
@@ -24,6 +24,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Collection;
@@ -39,6 +40,7 @@ import org.apache.bcel.classfile.JavaClass;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugin.logging.Log;
+import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.util.FileUtils;
 import org.codehaus.plexus.util.cli.Commandline;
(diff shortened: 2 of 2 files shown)


jcflack:
Hi,

I've glanced at the changes here, but I am not (at least not right away) seeing how they would help with #204. It won't hurt, and is probably good, but what was at issue in #204 was that `IncludePath` doesn't have a meaningful `toString` override, not that `Compiler` doesn't. In fact, as I look at some debug output here, it would be helpful if `IncludePath`, `C`, `Javah`, `Lib`, and `Linker` had meaningful overrides for `toString`:

```
[DEBUG] Configuring mojo 'com.github.maven-nar:nar-maven-plugin:3.2.3:nar-validate' with basic configurator -->
[DEBUG]   (f) baseDir = /home/foo/pljava/pljava-so
[DEBUG]   (f) options = [-Wno-sign-conversion]
[DEBUG]   (f) defines = [PLJAVA_SO_VERSION=1.5.0-BETA2]
[DEBUG]   (s)  = /home/foo/pljava/pljava-so/src/main/include/
[DEBUG]   (s)  = /home/foo/pljava/pljava-so/target/nar/javah-include/
[DEBUG]   (f) includePaths = [com.github.maven_nar.IncludePath@93c8c0, com.github.maven_nar.IncludePath@60de94, com.github.maven_nar.IncludePath@1b4d206, com.github.maven_nar.IncludePath@48a168]
[DEBUG]   (s) c = com.github.maven_nar.C@5cdf19
[DEBUG]   (f) classesDirectory = /home/foo/pljava/pljava-so/target/classes
[DEBUG]   (f) failOnError = true
[DEBUG]   (f) gnuSourceDirectory = /home/foo/pljava/pljava-so/src/gnu
[DEBUG]   (f) ignore = false
[DEBUG]   (f) classPaths = [/home/foo/pljava/pljava-so/../pljava/target/classes/, /home/foo/pljava/pljava-so/../pljava-api/target/classes/]
[DEBUG]   (f) classDirectory = /home/foo/pljava/pljava-so/../pljava/target/classes
[DEBUG]   (f) extraClasses = [java.sql.Types]
[DEBUG]   (f) javah = com.github.maven_nar.Javah@116b06b
[DEBUG]   (f) layout = com.github.maven_nar.NarLayout21
[DEBUG]   (f) type = plugin
[DEBUG]   (f) linkCPP = false
[DEBUG]   (f) libraries = [Library: type: plugin]
[DEBUG]   (f) libtool = false
[DEBUG]   (f) name = ecpg
[DEBUG]   (f) type = shared
[DEBUG]   (f) name = pgtypes
[DEBUG]   (f) type = shared
[DEBUG]   (f) name = pq
[DEBUG]   (f) type = shared
[DEBUG]   (f) libs = [com.github.maven_nar.Lib@1c94d46, com.github.maven_nar.Lib@f03ef7, com.github.maven_nar.Lib@b5f7bf]
[DEBUG]   (f) linker = com.github.maven_nar.Linker@9c1f03
[DEBUG]   (f) localRepository =        id: local
      url: file:///home/foo/.m2/repository/
   layout: none

[DEBUG]   (f) maxCores = 1
[DEBUG]   (f) onlySpecifiedCompilers = false
[DEBUG]   (f) outputDirectory = /home/foo/pljava/pljava-so/target
[DEBUG]   (f) remoteArtifactRepositories = [       id: central
      url: http://repo.maven.apache.org/maven2
   layout: default
snapshots: [enabled => false, update => daily]
 releases: [enabled => true, update => daily]
]
[DEBUG]   (f) runtime = dynamic
[DEBUG]   (f) skip = false
[DEBUG]   (f) mavenProject = MavenProject: org.postgresql:pljava-so:1.5.0-BETA2 @ /home/foo/pljava/pljava-so/pom.xml
[DEBUG] -- end configuration --
```
You can see those are the items that are showing up in the debug output as *classname*`@`*hexaddress* instead of showing their actual values. I don't know if `prettyMavenString` is what they need, because I'm not sure all of those classes *have* their values injected to annotated fields ... they might just be getting constructed by other things. Maybe those 5 classes just need simple `toString` methods added?

Some other objects, like those representing class paths, do seem to have them already.

-Chap

ctrueden:
OK @jcflack, if there are other classes which you feel would benefit from `toString()` methods, you are [very welcome](https://github.com/maven-nar/nar-maven-plugin/wiki/Frequently_Asked_Questions#q-is-the-nar-project-being-actively-developed) to submit a PR. This type of change is highly unlikely to affect backward compatibility, so getting it merged should not be too burdensome as long as your patches are clean.

ctrueden:
Merging, since it doesn't hurt anything. But it does not close #204, since I implemented `toString()` for the wrong class.

PR #214, opened 2016-03-03
MSVC Intermediate output files to AOL target folder

GregDomjan:
With change to build working dir, so that building from src folders rather than AOL nar/obj folder or nar-test/obj,  some additional settings needed to be made for msvc output locations.

Also I got tired of Windows delete failing for NarUnpack and leaving an empty folder thats been modified so that Unpack doesn't try again without having to manually go and delete from the narunpack folder.


ctrueden:
Awesome, @GregDomjan. I just offered some comments about DRYness ("Don't Repeat Yourself"). If you agree, feel free to push patches along those lines. Or if you are too busy, this PR already LTGM. :+1: Merge anytime you're ready.

ctrueden:
:+1:

PR #215, opened 2016-03-07
Implement libs.names TODO in pom.xml.

gwilga:
This features add the possibility to name the libraries that will be available within the nar package.
Usefull for linking process with dependencies and to package a project with multiple libraries.

Add the tag configuration \<libsName\>lib1,lib2\</libsName\> to specify usage of lib1 and lib2 in the package.


ctrueden:
Looks like a nice clean change.

However, before I merge, I have one naive question: would it make sense to use `List<String>` instead of a single `String` for the field, so that you specify in the standard Maven syntax:
```xml
<libNames>
   <libName>lib1</libName>
   <libName>lib2</libName>
</libNames>
```
?
I know this is more verbose, but it also leaves a little less room for error.

GregDomjan:
Was ambivalent when I saw this change.

This is kind of dirty from the Maven view of artifact being a single unit and having multiple units grouped together should be avoided, so I wonder at making this easier.

However I've also found in quickly packaging 3rd party binaries and coming to understand how to use maven, it can be helpful to list multiple libs to be included into the linker.  As I wasn't building I could put the lib names in the output name and cheat.

---

NAR seems to mix between delimited string like CSV and lists, sometimes providing both.
Lists can be merged, delimited strings can only be overwritten.  I can't see a reason that libNames should be inherited and merged as it's fairly specific to the artifact.  CSV is probably more appropriate from that.

Should doc changes be included, and perhaps sample/integration tests?

gwilga:
I have added one integration test (it0029), this one runs on linux platform. I did not enable it for windows.
I have change the documentation.
But I have not change the CSV delimited string behavior. (regarding GregDomjan comment).

By the way, I was aware that for the Maven view this change is not compliant. But I think it can save a lot of time when packaging 3rd party binaries.

GregDomjan:
I'd merge this, waiting on any input from @ctrueden

ctrueden:
Sorry for the delay. Recently returned from vacation.

Thanks for your efforts on this, @gwilga.

axel3rd:
Hi,

@ctrueden : Good luck for the return from vacation :-)

Is there a hope for a release (v3.3.1?) with this feature shortly (~week) ?

I don't want changing/influencing the roadmap, but this feature is important for our Cxx CI strategy deployment (discussed with @gwilga, we work together) ; it is just for knowing if we deploy this PR in our internal company repository with "intermediate version", or if we wait a little.

Thank you very much.
Best regards.

ctrueden:
There is not really a roadmap; we do a release whenever someone needs one.

So, last night, I released `3.4.0` for you. (Minor version bump according to SemVer, because new API was added since the last release.)

Enjoy! :smile:

axel3rd:
@ctrueden : Thank you very much ! :)

PR #218, opened 2016-03-15
Msvc

GregDomjan:
Was working on issues with "LINK : fatal error LNK1149: output filename matches input filename" f5a9d98  and identified several other fixes along the way.


ctrueden:
Patches look really good, thanks! My usual caveat applies: I cannot test easily on Windows, so merging early. Any issues arising can be addressed in later patches.

PR #220, opened 2016-03-30
attempt to solve issue 217 by introducing new version comparison

Oneplus:


src/main/java/com/github/maven_nar/Msvc.java
```diff
@@ -18,7 +18,6 @@
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.tools.ant.types.Environment.Variable;
-import org.codehaus.plexus.component.configurator.converters.basic.DoubleConverter;
```
ctrueden: Would be better to just squash this into the previous commit using `git rebase -i HEAD^^` and changing the second line from `pick` to `fixup`

Oneplus:
Hi @ctrueden , I've changed Double.compare to DefaultArtifactVersion.compareTo according to your suggestion and rebased the useless commit.

ctrueden:
Thank you!

PR #221, opened 2016-04-01
Cleancode

DavidZemon:
I just imported the project into IntelliJ 2016.1.1 and ran a full code inspection analysis. Found lots of issues and fixed just a few of the obvious ones that had a small likelihood of introducing bugs. Maven still passes the test and verify lifecycles, so it should be good-to-go still.

I avoided potentially dangerous fixes like converting `Vector` to `List` (likely okay, since the code was probably written before List existed) and `StringBuffer` to `StringBuilder`.

There are also lots of instances of Vector without its generic argument provided that I was going to fix, but the first one I tried appeared to contain two different types (String and File) so I didn't make any of those fixes either.

### DIFF
--- a/src/main/java/com/github/maven_nar/AbstractDependencyMojo.java
+++ b/src/main/java/com/github/maven_nar/AbstractDependencyMojo.java
@@ -22,14 +22,12 @@ package com.github.maven_nar;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.InputStream;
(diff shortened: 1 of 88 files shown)


ctrueden:
Thank you @DavidZemon!

FYI, your Git commits here do not match your GitHub username, due to mismatched email address.

DavidZemon:
Oh, thanks for letting me know. I didn't even think of that. Did that on my work computer.... Guess I better look into per-project email addresses in git

PR #227, opened 2016-04-29
Should now be able to handle spaces in filenames

SteveSmith16384:
I've not been able to fully test this, as I don't use Maven NAR directly myself (only as part of a shared project).  However, you might find the suggested changes useful.

### DIFF
--- a/src/main/java/com/github/maven_nar/NarUtil.java
+++ b/src/main/java/com/github/maven_nar/NarUtil.java
@@ -386,7 +386,7 @@ public final class NarUtil {
     if (file.isFile() && file.canRead() && file.canWrite() && !file.isHidden()) {
       // chmod +x file
       final int result = runCommand("chmod", new String[] {
-          "+x", file.getPath()
+          "+x", file.getPath().replaceAll(" ", "\\ ")
       }, null, null, log);
       if (result != 0) {
         throw new MojoExecutionException("Failed to execute 'chmod +x " + file.getPath() + "'" + " return code: \'"
--- a/src/main/java/com/github/maven_nar/NarGnuConfigureMojo.java
+++ b/src/main/java/com/github/maven_nar/NarGnuConfigureMojo.java
@@ -136,7 +136,7 @@ public class NarGnuConfigureMojo extends AbstractGnuMojo {
             getLog().info("Running GNU " + BUILDCONF);
             String gnuBuildconfArgsArray[] = null;
             if (this.gnuBuildconfArgs != null) {
-              gnuBuildconfArgsArray = this.gnuBuildconfArgs.split("\\s");
+              gnuBuildconfArgsArray = this.gnuBuildconfArgs.split("[^\\s\"']+|\"([^\"]*)\"|'([^']*)'");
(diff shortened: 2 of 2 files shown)


src/main/java/com/github/maven_nar/NarUtil.java
```diff
@@ -386,7 +386,7 @@ public static void makeExecutable(final File file, final Log log) throws MojoExe
     if (file.isFile() && file.canRead() && file.canWrite() && !file.isHidden()) {
       // chmod +x file
       final int result = runCommand("chmod", new String[] {
-          "+x", file.getPath()
+          "+x", file.getPath().replaceAll(" ", "\\ ")
```
GregDomjan: Because `chmod` is called directly rather than using shell `sh` additional escaping or quoting for spaces shouldn't be required.
Was this required or "added for safety" ?

GregDomjan:
Thanks @SteveSmith16384 

I'd really prefer to avoid managing quotes as it tends to get messier and messier as additional layers are added.   But as sh is used it seems it has to be managed.

Did a quick check of the regex used for splitting  `"[^\\s\"']+|\"[^\"]*\"|'[^']*'"`
Found  http://stackoverflow.com/questions/366202/regex-for-splitting-a-string-using-space-when-not-surrounded-by-single-or-double

The following comments where interesting, do we need to account for unmatched quotes?

- The problem with this answer is with unmatched quote: John's mother results splited in [John, s, mother] – leonbloy May 16 '14 at 20:10	
- To fix the issue leonbloy outlines, you can re-order the operands a bit and omit the quotes from the whitespace-group: `"([^"]*)"|'([^']*)'|[^\s]+`  – Ghostkeeper Sep 15 '14 at 1:26


Thinking we need to make this new version of splitting space separated values into a util function as there seem to be additional locations to potentially fix.
It would have been better I think to make some if not all of these inputs as list still instead of having to deal with issues of string splitting. 

```
Compiler.java
322: final String[] opts = this.optionSet.split("\\s");  
Linker.java
344: final String[] opts = this.optionSet.split("\\s");  
NarGnuConfigureMojo.java
139: gnuBuildconfArgsArray = this.gnuBuildconfArgs.split("\\s");  

Compiler.java (3 matches)
338: final String[] option = optionsProperty.split(" ");  
503: final String[] exclude = defaultExcludes.split(" ");  
548: final String[] include = defaultIncludes.split(" ");  
Linker.java
358: final String[] opt = option.split(" ");  
NarGnuConfigureMojo.java
157: final String[] a = this.gnuConfigureArgs.split(" ");  
NarGnuMakeMojo.java (2 matches)
74: args = this.gnuMakeArgs.split(" "); 
90: args = this.gnuMakeArgs.split(" "); 
NarIntegrationTestMojo.java
952: final List args = Arrays.asList(this.argLine.split(" "));  

```

src/main/java/com/github/maven_nar/NarUtil.java
SteveSmith16384: I seem to remember this being required.

Issue #230, opened 2016-05-27
static binding for multi type builds


manu4linux:
we have a nar build case where it builds for types executable and static. 
we see that for this case nar.properties file always has binding as `executable` and this causes downstream builds to fail as they can't link to a static lib.
`libs.binding=executable 
i386-Linux-gpp.libs.binding=executable` 

please see mailing list query on this at "https://groups.google.com/forum/#!topic/maven-nar/Ig9yc8LPwxA"

we have fixed this by hacking "maven_nar/NarLayout21.java" file "prepareNarInfo" to handle  Library.STATIC similar to Library.SHARED.

line 203: would look like "if ( (type.equals(Library.SHARED)) || (type.equals(Library.STATIC)) )" 

I have tested it on CentOs7.2 i386 vm and All NAR IT tests seem to pass except one. This test also fails for original code due to non resolution of static -lstdc++ & others..

[INFO] -------------------------------------------------
[INFO] Build Summary:
[INFO]   Passed: 25, Failed: 1, Errors: 0, Skipped: 0
[INFO] -------------------------------------------------
[ERROR] The following builds failed:
[ERROR] *  it0002-executable-static/pom.xml
[INFO] -------------------------------------------------

I will send a pull request for the same.. "https://github.com/maven-nar/nar-maven-plugin/pull/231"
please comment on it ...

PR #231, opened 2016-05-27
static binding not picked up when built for both executable and static libs

manu4linux:
This is push is as per #230  and https://groups.google.com/forum/#!topic/maven-nar/Ig9yc8LPwxA

I have tested it on CentOs7.2 i386 vm and All NAR IT tests seem to pass except one "it0002-executable-static". This test also fails for original code due to non resolution of static -lstdc++ & others..

[INFO] -------------------------------------------------
[INFO] Build Summary:
[INFO] Passed: 25, Failed: 1, Errors: 0, Skipped: 0
[INFO] -------------------------------------------------
[ERROR] The following builds failed:
[ERROR] * it0002-executable-static/pom.xml
[INFO] -------------------------------------------------

please comment..

Thanks
Mano

### DIFF
--- a/src/main/java/com/github/maven_nar/NarLayout21.java
+++ b/src/main/java/com/github/maven_nar/NarLayout21.java
@@ -199,8 +199,8 @@ public class NarLayout21 extends AbstractNarLayout {
           narInfo.setLibs(aol, mojo.getLibsName());
         }
 
-        // We prefer libs of shared & static to jni/executable/none,
-        if ( (type.equals(Library.SHARED)) || (type.equals(Library.STATIC)) ) // overwrite whatever we had
+        // We prefer shared to jni/executable/static/none,
+        if (type.equals(Library.SHARED)) // overwrite whatever we had
         {
           narInfo.setBinding(aol, type);
           narInfo.setBinding(null, type);
@@ -209,7 +209,15 @@ public class NarLayout21 extends AbstractNarLayout {
           // jni/executable/none.
           if (narInfo.getBinding(aol, null) == null) {
             narInfo.setBinding(aol, type);
+          } else {
+            //static lib is preferred over other remaining types; see #231
+            if (type.equals(Library.STATIC))
+            {
+              narInfo.setBinding(aol, type);
+              narInfo.setBinding(null, type);
+            }
           }
+
           if (narInfo.getBinding(null, null) == null) {
             narInfo.setBinding(null, type);
           }
--- a/src/main/java/com/github/maven_nar/NarLayout21.java
+++ b/src/main/java/com/github/maven_nar/NarLayout21.java
@@ -199,14 +199,14 @@ public class NarLayout21 extends AbstractNarLayout {
           narInfo.setLibs(aol, mojo.getLibsName());
         }
 
-        // We prefer shared to jni/executable/static/none,
-        if (type.equals(Library.SHARED)) // overwrite whatever we had
+        // We prefer libs of shared & static to jni/executable/none,
+        if ( (type.equals(Library.SHARED)) || (type.equals(Library.STATIC)) ) // overwrite whatever we had
         {
           narInfo.setBinding(aol, type);
           narInfo.setBinding(null, type);
         } else {
           // if the binding is already set, then don't write it for
-          // jni/executable/static/none.
+          // jni/executable/none.
           if (narInfo.getBinding(aol, null) == null) {
             narInfo.setBinding(aol, type);
           }


ctrueden:
I am afraid I do not have sufficient insight to know whether this is a good idea.

@GregDomjan This is your code, from 891b94c6. Do you recall your reasoning? Would this change be OK?

GregDomjan:
I hope this makes sense...

Its limited impact to builds with multiple libraries/types means the change should be OK for most.
Personally would have some issues where we have some SDKs as static and shared, so this wouldn't be my preferred final solution, I'd have to reconfigure some projects to add config. 

I think It's a fair direction to take that default binding of static should be used in preference to executable.  
It would be my preference that Shared is still the preferred default over static (rather than leaving it to build order - or in packaging by directory order)

Using Static in pref to !Shared could be added as an else condition at 212,215  first checking the binding was not shared then I would be happy. 

---

A simpler way of specifying the binding for each dependency would help to resolve, rather than having to spend so much effort on a best estimate default.

---
The problem is more generally that if a project has multiple library compilation types which one is NAR to pick as the "default" binding linkage by dependants (to reduce configuration of dependants).

Prior to 891b94c it was whatever built last, now it's shared or whatever built first.
With this adjustment it's shared or static whichever is built last, or others that built first.
It adds a little uncertainty.

On the dependent side there is some inconsistency, and differs to what I recall being default of shared, 

* for download - static is the default decision,if ```dependencyBindings``` is not configured, and if the dependency doesn't have a default.
https://github.com/maven-nar/nar-maven-plugin/blob/master/src/main/java/com/github/maven_nar/AbstractDependencyMojo.java#L228-254

* for compile lib/link - none is the default decision if the dependency doesn't have a default.
https://github.com/maven-nar/nar-maven-plugin/blob/master/src/main/java/com/github/maven_nar/NarCompileMojo.java#L281-287
Doesn't seem to use ```library/dependencyBindings``` configuration?

ctrueden:
@GregDomjan Thanks for the explanation!

> A simpler way of specifying the binding for each dependency would help to resolve, rather than having to spend so much effort on a best estimate default.

Makes sense! But it seems to me like this would need to become part of the classifier then? I don't see anywhere else in the dependency metadata where we could cram that info.

> Using Static in pref to !Shared could be added as an else condition at 212,215 first checking the binding was not shared then I would be happy.

That would be a really easy change, and if it satisfies all use cases, then my vote is to just hardcode this order as the priority for now. @manu4linux Can you make the change and force push this branch accordingly? If everyone agrees this is OK (for now), then we can merge it.

manu4linux:
Thanks for feedback @GregDomjan and @ctrueden . I will make an effort on the suggestion made and will make a push.

manu4linux:
@ctrueden  I made update to code as per the feedback from @GregDomjan. Hope my interpretation is correct.

> Using Static in pref to !Shared could be added as an else condition at 212,215 first checking the binding was not shared then I would be happy.

As before I have tested it on CentOs7.2 i386 vm and All NAR IT tests seem to pass except one "it0002-executable-static". This test fails for original code too due to non resolution of static -lstdc++ & others..

> [INFO] -------------------------------------------------
> [INFO] Build Summary:
> [INFO]   Passed: 25, Failed: 1, Errors: 0, Skipped: 0
> [INFO] -------------------------------------------------
> [ERROR] The following builds failed:
> [ERROR] *  it0002-executable-static/pom.xml
> [INFO] -------------------------------------------------
> [INFO] ------------------------------------------------------------------------
> [INFO] BUILD FAILURE
> [INFO] ------------------------------------------------------------------------
> [INFO] Total time: 5:36.894s
> [INFO] Finished at: Tue Jun 07 11:23:56 PDT 2016


Also it would great if we can have below strategy as explained by GregDomjan as a next feature. In essence it would be Ideal if we can explicitly specify how each dependency would be resolved as.
> A simpler way of specifying the binding for each dependency would help to resolve, rather than having to spend so much effort on a best estimate default.

ctrueden:
> Also it would be great if we can have below strategy as explained by GregDomjan in next release. In essence it would be Ideal if we can explicitly specify how each dependency would be resolved as.

As I said, the only easy way I see of doing it would be to add that information to the classifier. So we'd have AOLB instead of just AOL. Does that make sense to you @GregDomjan ? This would be a breaking change from previous versions of NAR, so I'd want to see a more thorough review process than we have been using for the PRs over the past couple of years.

ctrueden:
@manu4linux Thanks. I pushed a cleaned up version of your patches as 43338ce6f14be779854d9c19ffc6f626fa83ef74.

manu4linux:
Thanks @ctrueden !

Issue #234, opened 2016-06-06
VisualStudio required


mr-raphael-martin:
Using NAR Plugin version 3.4.0 on a Windows 7 platform **where Visual Studio has never been installed**, Nar plugin always try to run the command "cmd.exe /X /C "link /?", i.e. try to check if MSVC++ is installed.
So it seems impossible to work with MINGW/GCC
You can try to force, platform, os, linker name or directly AOL in your POM, all these options are ignored, in the early goal VALIDATE.
To solve the problem, I had to get back to NAR 3.2.3 version. What a pity... I am quite desappointed because I used NAR plugin for years now...

ctrueden:
This is definitely a bug, not intended behavior. A PR fixing it would be most warmly welcomed. If you do not have time, then there is nothing wrong with continuing to use 3.2.3 for now.

mr-raphael-martin:
Hi Curtis
Thanks for your quick answer.
I will try to have a deeper look inside the code. I isolated already that the problem come from the initialization of the Nar mojos, all derivated from the same abstract Nar mojo. Unfortunately in the private method checking for MSVC default parameters, the case of not having any MSVC installed is not (correctly)implemented.
I will keep on investigating...
Le 10 juin 2016 23:07, Curtis Rueden <notifications@github.com> a écrit :This is definitely a bug, not intended behavior. A PR fixing it would be most warmly welcomed. If you do not have time, then there is nothing wrong with continuing to use 3.2.3 for now.

dugilos:
Hello all,
After a long pause I recently updated the plugin for my projects and I ran on the same error, so I corrected it and sent the pull request #238

ctrueden:
I assume the merge of #238 fixes this issue. If not, please reopen.

PR #238, opened 2016-06-18
Correction for mingw on Windows.

dugilos:
nar-maven-plugin doesn't work anymore on Windows with mingw **if
VisualStudio isn't installed**, because it tries to run the msvc command
"link" during the initialization of the class Msvc in
AbstractNarMojo.validate() (even if we don't use msvc).

Correction : in Msvc.init() I make the same test
"msvc".equalsIgnoreCase(mojo.getLinker().getName()) than in other
methods of the class (configureCCTask and configureLinker for example).
This way if the name of the linker if forced to "g++" in the pom (for
mingw), Msvc.init() doesn't do anything as if we were not on Windows and
thus doesn't quit on error.

### DIFF
--- a/src/main/java/com/github/maven_nar/Msvc.java
+++ b/src/main/java/com/github/maven_nar/Msvc.java
@@ -191,7 +191,7 @@ public class Msvc {
 
   private void init() throws MojoFailureException, MojoExecutionException {
     final String mojoOs = this.mojo.getOS();
-    if (NarUtil.isWindows() && OS.WINDOWS.equals(mojoOs)) {
+    if (NarUtil.isWindows() && OS.WINDOWS.equals(mojoOs) && "msvc".equalsIgnoreCase(mojo.getLinker().getName())) {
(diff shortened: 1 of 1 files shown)


markjohndoyle:
Hi dugilos,

What version is this? I did a quick test on our project with 3.5.0 and it worked on Windows with mingw.

Did we miss something in this commit?

https://github.com/maven-nar/nar-maven-plugin/commit/0bf008d0819d5cbd5fc393b6e1cc9ff49d72c708

dugilos:
Hi,

Sorry I didn't give enough details, the bug appears if VisualStudio isn't installed on the computer, as described in the issue #234 .

I ran my tests with the release version 3.5.0.

markjohndoyle:
Ah ok, that makes more sense as I installed a few weeks ago for node.

Looks like you did actually reference the issue - I just missed it completely!

dugilos:
I edited my first comment on top to be more precise for this pull request :)

PR #241, opened 2016-08-07
GCC WindresResourceCompiler, bad output folder for objects.

dugilos:
The method getInputFileArgument of the Compiler classes returns the
arguments to pass to the compiler based on the input files. On
WindresResourceCompiler the resulting arguments are
`-o<resource_filename_wo_ext>.<hash>.o <resource_fullpath>` :
* `<resource_filename_wo_ext>` = resource filename without path and
without extension.
* `<resource_fullpath>` = resource filename with its full path.

The object file is generated in `src/main/` and the linker exits then on error because it doesn't find the object file in the directory `target/nar/obj/<aol>/`

The method WindresResourceCompiler.getInputFileArgument is modified to
generate the command arguments
`-o<obj_fullpath>.<hash>.o <resource_relativepath>` :
* `<obj_fullpath>` : filename with full path for object files.
* `<resource_relativepath>` : resource filename with its relative path.

I took example on GccCompatibleCCompiler.getInputFileArgument.

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/gcc/WindresResourceCompiler.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/gcc/WindresResourceCompiler.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.util.Vector;
 
 import org.apache.tools.ant.types.Environment;
+import org.apache.tools.ant.util.FileUtils;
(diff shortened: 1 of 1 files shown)


ctrueden:
Thanks, @dugilos!

PR #245, opened 2016-08-24
#244 : full path to Source filenames added only to gcc builds.

### LINKED ISSUE #244: option of full path to filenames added for gcc builds
we see that all the debug string filenames with in executable now only have relative path to main. for example `/base/project/src/main/c/somefile.c` would show as `c/somefile.c`
Since our project use a static dependency libs, let say file from `/base/project2/src/main/c/someotherfile.c` will resolve to `c/someotherfile.c` causing GDB debugging issues. 

We have 14 million lines of code spread out all over the place. Our build agents end up with paths like this:

```
/usr/local/teamcity/buildagent/work/41ae010f80e2cbba/c/cs_timer.c, 
/usr/local/teamcity/buildagent/work/41ae010f80e2cbba/c/cs_sequence.c
```

for files physically at 
/home/user/server/modules/c-libs/**cs_timer/src/main**/c/cs_timer.c
/home/user/server/modules/c-libs/**cs_sequence/src/main**/c/cs_sequence.c

In our .gdbinit file, we can map `/usr/local/teamcity/buildagent/work/41ae010f80e2cbba/` to the top level of our source (as in `set substitute-path /usr/local/teamcity/buildagent/work/41ae010f80e2cbba/ /home/user/server/modules/c-libs/cs_timer/src/main`) and gdb can find the sources. But, without the full path and just `c/cs_env.c` or `c/cs_super.c` we’d have to have individual mapping for each file. That’s unworkable for us.

if I am not wrong Behavior in old `maven-nar-plugin`  seem to give full path which made it easy for GDB to find them.

FYI, pls see https://groups.google.com/forum/#!topic/maven-nar/sIBkEOdjDfA

manu4linux:
This pull request is to fix #244 issue of having full path to filenames in gcc builds with a new NAR parameter.

Main code change is to use full path in `getInputFileArgument`  based on the `gccFileAbsolutePath`  flag .

I have only added logic to`GccCompatibleCCompiler.java` as this fix is intended only for gcc. 
Same fix can be applied to other compilers too by adding logic to corresponding files below. But for me I don't have a way to test them.

```
borland/BorlandCCompiler.java
borland/BorlandResourceCompiler.java
gcc/WindresResourceCompiler.java
mozilla/XpidlCompiler.java
msvc/MsvcCompatibleCCompiler.java
msvc/MsvcMessageCompiler.java
msvc/MsvcMIDLCompiler.java
msvc/MsvcResourceCompiler.java
trolltech/MetaObjectCompiler.java
trolltech/UserInterfaceCompiler.java
```

This issue is also related to  #205  and to #158 issues.

To set `gccFileAbsolutePath`  flag `true` you need below NAR configuration pom update.

``` xml
        <cpp combine.children="append">
         <gccFileAbsolutePath>true</gccFileAbsolutePath>
         </cpp>
         <c combine.children="append">
         <gccFileAbsolutePath>true</gccFileAbsolutePath>
         </c>
```

This fix behaves exactly as before until `gccFileAbsolutePath`  flag is set to `true`. 

I have tested with `mvn clean install -Prun-its` for both `gccFileAbsolutePath` true & false and seem to be fine on my VM which is `Centos 7 x86 32bit`

### DIFF
--- a/src/main/java/com/github/maven_nar/Compiler.java
+++ b/src/main/java/com/github/maven_nar/Compiler.java
@@ -86,6 +86,16 @@ public abstract class Compiler {
   private File testSourceDirectory;
 
   /**
+   * To use full path for the filenames.
+   * false to have "relative" path
+   * true to have "absolute" path
+   * absolute: will give path from filesystem root "/"
+   * relative: will give relative path from "workdir" which is usually after "${basedir}/src/main"
+   */
+  @Parameter(required = true)
+  private boolean gccFileAbsolutePath = false;
+
+  /**
    * Include patterns for sources
    */
   @Parameter(required = true)
@@ -483,6 +493,9 @@ public abstract class Compiler {
         compilerDef.setWorkDir(this.sourceDirectory);
       }
     }
+
+    compilerDef.setGccFileAbsolutePath(this.gccFileAbsolutePath);
+
     return compilerDef;
   }
 
--- a/src/main/java/com/github/maven_nar/cpptasks/CCTask.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/CCTask.java
@@ -1232,6 +1232,7 @@ public class CCTask extends Task {
         CommandLineCompilerConfiguration commandLineConfig = (CommandLineCompilerConfiguration) config;
         AbstractCompiler compiler = (AbstractCompiler) commandLineConfig.getCompiler();
         compiler.setWorkDir(currentCompilerDef.getWorkDir());
+        compiler.setGccFileAbsolutePath(currentCompilerDef.getGccFileAbsolutePath());
         ProcessorConfiguration[] localConfigs = new ProcessorConfiguration[] {
           config
         };
--- a/src/main/java/com/github/maven_nar/cpptasks/CompilerDef.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/CompilerDef.java
@@ -61,6 +61,7 @@ public final class CompilerDef extends ProcessorDef {
   private String toolPath;
   private String compilerPrefix;
   private File workDir;
+  private boolean gccFileAbsolutePath;
 
   private boolean clearDefaultOptions;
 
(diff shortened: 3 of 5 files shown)


manu4linux:
@dmitrykirsanov @ctrueden Please review. 
Fix is based on @dmitrykirsanov comments on https://groups.google.com/forum/#!topic/maven-nar/sIBkEOdjDfA

dproteus13:
I'll admit I'm checking this quickly on my phone in line... but it looks like this changes the default behaviour to relative paths?  Could we just as easily default to absolute so that we notice no impact unless we want the change?

manu4linux:
Hi @dproteus13, thanks for  your time on review.
Yes, code does default to relative path as it was the same behaviour before this patch and i did not want to break anything when flag/parameter is not set.

Also as per cobersation in NAR groups https://groups.google.com/forum/#!topic/maven-nar/sIBkEOdjDfA  there seem to be some resoning went in to for having relative path.

At this point i am not sure which one is best for default behaviour, but I do see strong need of at least having an option for gcc builds to have full path filename symbols while debugging the code with GDB.

manu4linux:
@dmitrykirsanov request for quick comment on this pull, please!

dmitrykirsanov:
I don't mind if absolute path will be default behaviour and relative will be used by setting a parameter. Seems, it is more familiar and comfortable for maven-nar community, who use this plugin for long time.

manu4linux:
Thanks @dmitrykirsanov i do agree and I can do it for Gcc compiler to have default full path, but unfortunately I can't do the necessary changes to reflect all compilers.  Which puts me in situation where I feel current pull with targeted gcc compiler having switch is ideal for now. 

I will update #244 issue to have an future implementation request with default behaviour of full path for all compilers, such that we don't forget the conclusion we had here.

Let us know if you agree and that @ctrueden can merge this pull, please.

dmitrykirsanov:
@manu4linux, if you are asking my consent, I do not have objections. It is not difficult for us to update configuration a bit for using this change.

ctrueden:
Sorry for the long delay, everyone. I was seriously injured and am still recovering—it will be quite some time before I catch up on everything.

If anyone is keen to take over as primary maintainer of NAR, it would certainly help me out, and help this project to move forward more quickly.

In the meantime, since everyone seems agreed: I merge the PR! :smile:

PR #249, opened 2016-11-04
Fix NPE problem in isMSVC

yuxuanchiadm:
We need initialization linker first. So getLinker() will not return null in Msvc.isMSVC().


ctrueden:
Thanks! And sorry for the long delay.

Issue #122, opened 2014-07-19
Reloading, or injecting new, aol.properties in modules of a parent POM


abiggerhammer:
I'm responsible for a couple of legacy C++ projects with their own ideas about what shared library extensions ought to be, and for legacy reasons those ideas don't get to change. I'm able to get around this with nar-maven-plugin by creating an aol.properties file for each project to define an appropriate value for, in this case, amd64.Linux.gpp.shared.extension. However, when I try to build the two projects as child modules of a parent project, only one aol.properties file is ever read (that of the first child module built); the second module reuses the properties from the first one. Even if I explicitly set a `nar.aolProperties` property in each child POM to point to the file each one needs, that setting is not honored in the second child module.

I notice that in NarProperties.java there's a static method, `.inject()`, that is meant to overwrite existing AOL properties, but it doesn't appear like this method is used in any mojo. Am I missing something?

ctrueden:
@abiggerhammer With the merge of #251 thanks to @hboutemy, it should now be possible to override `aol.properties` the way you wanted. Please follow up here if you still have time and interest!

ctrueden:
Presuming fixed. Please reopen if not.

PR #251, opened 2016-11-20
Add support for aol.properties inheritance/override #122

hboutemy:
here is an enhancement permitting inheritance of aol.properties from parent project or override


ctrueden:
Thank you @hboutemy! Beautiful commit. 😁

PR #253, opened 2017-01-18
Fix NPE because SystemRoot was not Initialized properly

eyalg1972:
The AbstractNarMojo class is initializing the MSVC object before the linker is defined to be "msvc".
As a result- the call to the init() method does not pass the linker condition and the Windows variables are not being set -  windowsHome is not set to "%SystemRoot%".

The fix is that if the setMojo method is called after the compiler is being defined, the linker is "msvc" but the this.windowsHome was not set-> it will call the init() method


eyalg1972:
Sorry- did not saw that in the latest sources it was already fixed, you can ignore it

PR #255, opened 2017-01-22
Support vs2005

eyalg1972:
Currently with the separation between Visual Studio and Windows SDK-> Visual Studio 2005 can not be used with the NAR Plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -297,7 +297,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>2.3.2</version>
+          <version>3.1</version>
         </plugin>
         <plugin>
           <artifactId>maven-site-plugin</artifactId>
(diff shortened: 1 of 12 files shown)


ctrueden:
@GregDomjan Any concerns with this?

GregDomjan:
I guess I commented on this chain in another pull request.
"Add Support to VS 2005" is really about identifying the SDK that comes with that version, and how to specify it.  At the moment based on linker version it is forcing to specific SDK.
If you use 2005 and a standalone SDK then this change would ignore the choice of standalone SDK.

eyalg1972:
Hi,

THe problem is that the current implementation is not allowing using the VS
2005 internal SDK, but automatically search for the new SDK, and as a
result I could not compile the code with the VS 2005 SDK.

We can try to check, if someone configured the SDK version- then use it,
but if not- to use the VS 2005 SDK (Not like in the newer VS version that
it search for the latest SDK that is installed).

What do you think?
Eyal

ctrueden:
@eyalg1972 It looks like this PR duplicates commits from other PRs.

Could you please rebase over the latest master, then force push? It would be nice if this PR was clean, so we could see what it is actually doing (as opposed to what other PRs are doing).

ctrueden:
Closing in favor of #269.

PR #257, opened 2017-01-22
Add definiton and options that are mandatory for Linux (And optional for Windows)

eyalg1972:
Linux require the -m32/-m64 compilation flags for compilation, so the NAR plugin should have them by default

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -297,7 +297,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>2.3.2</version>
+          <version>3.1</version>
         </plugin>
         <plugin>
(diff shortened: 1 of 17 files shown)


GregDomjan:
The pull request description for this set of patches doesn't really match up.

The changes for VS 2005/2008  included SDK support based on the version of VS forcing the platform sdk to only use the one provided by VS is not so good.  We used to compile using VS2005 with SDK 7.0
I guess the problem is how to identify that SDK that comes with.

Using https://en.wikipedia.org/wiki/Microsoft_Windows_SDK for reference,  the platform sdk included was probably 5.2 or 5.1.  The Visual Studio 'included' platform sdk all now have 'a' in version.

eyalg1972:
I fixed this Pull request.

It is about adding just few definitions to the aol.properties

Eyal

Issue #225, opened 2016-04-21
Nar breaks if there is a space in the path to the source


SteveSmith16384:
If there is a space in the path to the source, then the build breaks quite soon with something like:-

[INFO] --- nar-maven-plugin:3.2.0:nar-gnu-configure (default-nar-gnu-configure) @ TCSFirmware ---
[INFO] Copying GNU sources
[INFO] Running GNU configure
[INFO] + chmod +x /media/stephen/SSD Drive/code/xia/firmware/target/nar/gnu/amd64-Linux-gpp/src/configure
[INFO] args: [/media/stephen/SSD Drive/code/xia/firmware/target/nar/gnu/amd64-Linux-gpp/src/configure, --prefix=/media/stephen/SSD Drive/code/xia/firmware/target/nar/gnu/amd64-Linux-gpp/target, --prefix=/media/stephen/SSD, Drive/code/xia/firmware/target/hex]
[INFO] + cd /media/stephen/SSD Drive/code/xia/firmware/target/nar/gnu/amd64-Linux-gpp/src
[INFO] + sh /media/stephen/SSD Drive/code/xia/firmware/target/nar/gnu/amd64-Linux-gpp/src/configure --prefix=/media/stephen/SSD Drive/code/xia/firmware/target/nar/gnu/amd64-Linux-gpp/target --prefix=/media/stephen/SSD Drive/code/xia/firmware/target/hex
[ERROR] /media/stephen/SSD Drive/code/xia/firmware/target/nar/gnu/amd64-Linux-gpp/src/configure: 12: test: x--prefix=/media/stephen/SSD: unexpected operator

.. which leads to lots of further errors.  This seems to be cause by the function MakeExecutable() in NarUtils.jar (among other places) not wrapping the path in apostrophes when it runs the chmod command.

ctrueden:
Thanks for the report, @SteveSmith16384. Care to file a PR to address it? It would be super helpful.

GregDomjan:
It looks to me like chmod worked successfully.  Otherwise should see `NarUtil.makeExecutable` raise 
`MojoExecutionException("Failed to execute 'chmod +x " + file.getPath() + "'" + " return code: \'"
            + result + "\'.");`

This seems like it might be an issue in using "sh" (shell?) to call configure rather than calling configure directly.   "sh" is used in body of `NarGnuConfigureMojo.narExecute()` and  `runAutogen`

[...]

Issue #223, opened 2016-04-11
Integration tests fail on Windows


ctrueden:
Ever since the merge of #170, the integration tests fail on Windows. This needs to be resolved to avoid Windows-specific regressions in the code in the future.

Oneplus:
all integration tests passed on my own laptop which has only **one** msvc installation. But my CI failed on a system that has **multiple** msvc installations. https://ci.appveyor.com/project/Oneplus/ltp4j/build/14. and I think the problem points to #217 . Could this be the same reason for integration tests failures?

ctrueden:
Thanks for the additional information, @Oneplus. For what it's worth, the job configuration to which I refer is [here](http://jenkins.imagej.net/job/NAR-Maven-Plugin-windows/). Ultimately, it would be good to migrate that CI build off of that server and onto a more community-facing one such as Appveyor and/or Travis CI.

Unfortunately, I don't have any time to pursue that, nor to diagnose/fix the IT failures on (some) Windows systems. But if anyone else wants to take the lead on it, I will do everything I can to help.

Oneplus:
Hi @ctrueden If you are considering put the CI to Appveyor and Travis CI, it's pretty easy. All you need to do is to add the project to the CI service and add CI configuration file to the project.

I've added my own fork of nar-maven-plugin to the Appveyor and Travis CI and here is the IT result.

* https://ci.appveyor.com/project/Oneplus/nar-maven-plugin
* https://travis-ci.org/Oneplus/nar-maven-plugin/jobs/122403758

the windows and linux IT pass. However, the OSX IT fails and it says

```
[INFO] -------------------------------------------------
[INFO] Build Summary:
[INFO]   Passed: 22, Failed: 2, Errors: 0, Skipped: 0
[INFO] -------------------------------------------------
[ERROR] The following builds failed:
[ERROR] *  it0026-native-lib-loader/pom.xml
[ERROR] *  it0028-scope/pom.xml
```

Unfortunately, I am not an Apple users and cannot work on this issues.

Anyway, If you want to use Appveyor and Travis CI, you can merge my PR #220 and add the project in the CI services (it requires the maintainer of the project).

Oneplus:
One more thing to note is that PR #220 also contains a fix for the MSVC compiler version comparison problem as #217 .

karypid:
Hi,

It looks like something is already broken for windows:

ctrueden:
[...]

Issue #41, opened 2013-11-07
Some integration tests fail on OS X 10.9 and/or Java7


ctrueden:
## Linker issues

target/it/it0007-lib-shared/build.log:

```
[INFO] Linking...
[ERROR] Undefined symbols for architecture x86_64:
[ERROR]   "_main", referenced from:
[ERROR]      implicit entry/start for main executable
[ERROR] ld: symbol(s) not found for architecture x86_64
[ERROR] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

target/it/it0010-lib-static/build.log:

```
[INFO] Linking...
[ERROR] Undefined symbols for architecture x86_64:
[ERROR]   "_main", referenced from:
[ERROR]      implicit entry/start for main executable
[ERROR] ld: symbol(s) not found for architecture x86_64
[ERROR] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

target/it/it0014-multi-module/build.log:

```
[INFO] Linking...
[ERROR] Undefined symbols for architecture x86_64:
[ERROR]   "_main", referenced from:
[ERROR]      implicit entry/start for main executable
[ERROR] ld: symbol(s) not found for architecture x86_64
[ERROR] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

target/it/it0016-layout/build.log:

```
[INFO] Linking...
[ERROR] Undefined symbols for architecture x86_64:
[ERROR]   "_main", referenced from:
[ERROR]      implicit entry/start for main executable
[ERROR] ld: symbol(s) not found for architecture x86_64
[ERROR] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

target/it/it0018-fortran/build.log:

```
[INFO] Linking...
[ERROR] ld: library not found for -lgfortran
[ERROR] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

target/it/it0023-local-aol-properties/build.log:

```
[INFO] Linking...
[ERROR] Undefined symbols for architecture x86_64:
[ERROR]   "_main", referenced from:
[ERROR]      implicit entry/start for main executable
[ERROR] ld: symbol(s) not found for architecture x86_64
[ERROR] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
## Java configuration (looking for `javah` within the JRE)

target/it/it0003-jni/build.log:

```
[...]

PR #258, opened 2017-01-22
Improve implementation of executing command

eyalg1972:
Instead of using the "Execute" class of ANT, I am now using the standard Java execution process using ProcessBuilder.  Also- the executed command is printed so it is easy to see what might fail

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/CUtil.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/CUtil.java
@@ -400,7 +400,8 @@ public class CUtil {
       final boolean newEnvironment, final Environment env) throws BuildException {
     try {
       task.log(Commandline.toString(cmdline), task.getCommandLogLevel());
-      final Execute exe = new Execute(new LogStreamHandler(task, Project.MSG_INFO, Project.MSG_ERR));
+
+     /* final Execute exe = new Execute(new LogStreamHandler(task, Project.MSG_INFO, Project.MSG_ERR));
       if (System.getProperty("os.name").equals("OS/390")) {
         exe.setVMLauncher(false);
       }
@@ -418,6 +419,8 @@ public class CUtil {
       }
       exe.setNewenvironment(newEnvironment);
       return exe.execute();
+            */
+	  return CommandExecution.runCommand(cmdline,workingDir,task);
     } catch (final java.io.IOException exc) {
       throw new BuildException("Could not launch " + cmdline[0] + ": " + exc, task.getLocation());
     }
--- a/src/main/java/com/github/maven_nar/cpptasks/CommandExecution.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/CommandExecution.java
@@ -1,0 +1,93 @@
+package com.github.maven_nar.cpptasks;
+
+
+import org.apache.tools.ant.Project;
+
+import java.io.*;
+
+
+class StreamGobbler extends Thread {
+    InputStream is;
+    String type;
+    CCTask task;
+
+
+    StreamGobbler(InputStream is, String type, CCTask task) {
+        this.is = is;
+        this.type = type;
+        this.task = task;
+    }
+
+
+    public void run() {
+        try {
+            InputStreamReader isr = new InputStreamReader(is);
+            BufferedReader br = new BufferedReader(isr);
+            String line;
+            while ((line = br.readLine()) != null)
+                task.log(type +">"+ line );
+
+        } catch (IOException ioe) {
+            ioe.printStackTrace();
+        }
+
+
+    }
+}
+
+
+public class CommandExecution {
+
+
+
+
+    public static int runCommand(String[] cmdArgs, File workDir, CCTask task) throws  IOException{
+
+
+        try {
+
+            //Create ProcessBuilder with the command arguments
+            ProcessBuilder pb = new ProcessBuilder(cmdArgs);
+
+            //Redirect the stderr to the stdout
+            pb.redirectErrorStream(true);
+
+            pb.directory(workDir);
+
+            //Start the new process
+            Process process = pb.start();
+
+
+            // Adding to log the command
+            StringBuilder builder = new StringBuilder();
+            for(String s : cmdArgs) {
+
+                builder.append(s);
+                //Append space
+                builder.append(" ");
+            }
+            task.log("Executing - " + builder.toString(), Project.MSG_INFO);
+
+
+            //Create the StreamGobbler to read the process output
+            StreamGobbler outputGobbler = new StreamGobbler(process.getInputStream(), "OUTPUT",task);
+
+            outputGobbler.start();
+
+            int exit_value;
+
+            //Wait for the process to finish
+            exit_value = process.waitFor();
+
+
+            return exit_value;
+
+        } catch (InterruptedException e) {
+
+            e.printStackTrace();
+            return -2;
+        }
+
+    }
+
+}


ctrueden:
In principle I am all for simplifying the command execution. However, this is a deep change which needs to be tested on several platforms. It makes me nervous to merge it without some community validation first.

Ideally we would have all the integration tests passing on all of the major platforms, so that we can verify that they all still work after this change. Unfortunately, this is not the current reality—some integration tests fail on some platforms, for reasons no one has had time to investigate.

May I ask why you felt this change is necessary? Does it fix specific problems you encountered? If so, would you be willing to write a new test which fails under the old execution method, but passes with this one?

eyalg1972:
Hi,

I actually did here 2 things:

1) Replaced the ANT implementation- which I am not sure how it is
implemented with the newest implementation (Newest- from Java 1.5)
Since the NAR compilation is now done with Java 1.7, and Java 1.7 update 21
(If I am not mistaken) had modified the old way the Java used of
Runtime.exec, I prefered to be sure we have the recommendded way of
activating the commands (If the compilation was still with Java 1.5 or 1.6
it was less urgent).
2) Added to the log the command that is being executed, so in case of
failure we will have the full command already in the logs.

I can not remember now if there was a failure or not in any command with
the old implementation, but I did test the new implementaion on 4 platforms
and it works-

   - Windows 64 bits with MSVC
   - Linux 64 bits with g++
   - AIX 64 bits with g++
   - Solaris 64 bits with g++
   - AIX 64 bith with XLC_r will be tested in few weeks.

Regards

Eyal

eyalg1972:
Regarding-

Ideally we would have all the integration tests passing on all of the major
platforms, so that we can verify that they all still work after this
change. Unfortunately, this is not the current reality—some integration
tests fail on some platforms, for reasons no one has had time to
investigate.

Can you tell me-

1) How do you run today the integration tests?
2) What tests fails and on which platforms?
3) Can I re-produce the failures?

I am not promising, but I can try to help here.

Eyal

ctrueden:
> How do you run today the integration tests?

https://github.com/maven-nar/nar-maven-plugin/wiki/Testing

> 2) What tests fails and on which platforms?

On Windows, there is an issue: #223. A bit out of date now, though. Haven't tested lately.

On macOS, there have been issues in the past which were fixed (see #41), but now there are failures again, at least on my system. I filed new issue #267 documenting which ones.

> 3) Can I re-produce the failures?

I dunno! Try `mvn -Prun-its` and see what happens on your box! 😄 

> I am not promising, but I can try to help here.

I completely understand. No worries. Any progress you make is appreciated, but I have no expectations.

eyalg1972:
Can I get example of the failures on both Platforms- Windows and MacOS?

I mean- the exact errors that you get on your environment?

Eyal

eyalg1972:
I am running the Integration tests on my machine (WIndows 10, Java 1.7 and
3 different VS installed) and it pass.

Attaching log.

What am I not doing correct?

I will try to run it also on Mac in the next few days.

Eyal

ctrueden:
> What am I not doing correct?

Probably, you are doing everything correctly. It's just that there is a lot of variation between environments. Our old Windows node which was failing ran Windows 7, likely with different version(s) of Visual Studio. It is a lot of work to test a wide variety of configurations.

eyalg1972:
Can you attach me for the first failure-  The target folder?  so I can see
what is causing the error?

I already wrote you in the Mac OS issue that I found the source of the
problem-  Next week I will try to see if my proposed solution  is working.

Eyal

בתאריך 20 באפר׳ 2017 21:35,‏ "Curtis Rueden" <notifications@github.com> כתב:

ctrueden:
I pushed all IT logs from that attempt to the same gist. [Here is `it0001-executable.log`](https://gist.github.com/ctrueden/551535b1e1fdba5bcd499ff0e2035afe#file-it0001-executable-log).

eyalg1972:
Thanks Curtis,

However- this are the results on the Mac machine.
I already re-produce it and I believe that the problem is that the Mac now
does not pass the DYLD_LIBRARY_PATH to the surefire plugin process, so it
does not find the jnilib.
I suhhest to replace the NarSystem implementation- instead of calling
System.LoadLibrary with the module name (Which depends on
java.library.path), to use System.load which recieve full path.
I will test this locally on my machine next week.
[ Assuming it will work, I will try to do the change on seperate branch, do
we have somewhere in the code when we reference to the full path of the
artifact that was created?]

Regarding Windows failures-
1) Can you tell me- on that machine with Windows 7- Which VS are installed?
which Java version is used?
2) Can you send me the content of the target folder of the first failures?
I need only the build.log but the entire target folder to examine the
failure (FOr the MAC for example, I needed the surefile-reports folder)

Eyal

ctrueden:
> Regarding Windows failures-
> 1) Can you tell me- on that machine with Windows 7- Which VS are installed?
> which Java version is used?

I'm sorry, I don't have that information anymore. The VM in question was probably infected with malware, so we shut it down permanently. And the person who set it up has left the project; I do not know its configuration specifically.

If your Windows environment has all passing tests, then that is great. I wouldn't worry too much about it. The main thing is that we need to get Appveyor builds working properly, which means creating a maven-nar organization Appveyor account.

eyalg1972:
Can you clarify what do you mean by "The main thing is that we need to get
Appveyor builds working properly, which means creating a maven-nar
organization Appveyor account" ?
Who can create it and what is needed for it?
How do we run the tests on MacOS and Linux?

Also- Can you approve the pull request? (I hope in a week or two create a
fix for the MacOS problem)

Basically it is replacement of Runtime.exec with process builder, than
since Java 7 update 21 is recommennded (Actually- it is recommended since
Java 1.5, but in 1.7 update 21 they made some changes), I add few issues
with Runtime.exec in the past and always move to process builder
helped....  (In this case- it gives me also the option of printing to the
logs the command executed, which makes my life much easier, without the
need to run -X)

http://www.oracle.com/technetwork/java/javase/7u21-relnotes-1932873.html#jruntime
"Alternatively, the preferred way to create operating systems processes
since JDK 5.0 is using java.lang.ProcessBuilder. The ProcessBuilder class
has a much more complete API for setting the environment, working directory
and redirecting streams for the process."

It might resolve-  https://github.com/maven-nar/nar-maven-plugin/issues/225


Eyal

eyalg1972:
Sorry- NOw I see that 225 is related to a different execution command.

A diffrent story (But even there- we can replace the makeExecutable
implementation with Java 1.7 implementation for setting the execute
permissions)

Eyal

ctrueden:
> Can you clarify what do you mean by "The main thing is that we need to get Appveyor builds working properly, which means creating a maven-nar organization Appveyor account" ?

See #275.

eyalg1972:
And what is required to create such account?

Since we are open source project? (I assume you want organizational
account)?

Eyal

ctrueden:
A couple of updates on the CI: 1) I finally created the `maven-nar` AppVeyor account; see [the project page](https://ci.appveyor.com/project/maven-nar/nar-maven-plugin). The configuration needs work, though. 2) I activated the Travis CI build as well; see [here](https://travis-ci.org/maven-nar/nar-maven-plugin). So now it should be doable to configure these systems to test PRs more thoroughly cross-platform.

And a question: for executing processes, would it make sense to lean on [Apache Commons Exec](https://commons.apache.org/proper/commons-exec/)?

eyalg1972:
The Apache Commons Exec still rely on Runtime.getRuntime().exec, If only it
was using the ProcessBuilder instead...

Eyal

ctrueden:
Really sorry for my long silence on this PR, @eyalg1972.

I am going to merge it, and we'll see how it goes with Travis CI.

The AppVeyor CI still needs to be fixed, but that is definitely a separate issue from this change.

eyalg1972:
Thanks!!

Eyal

PR #259, opened 2017-01-22
Upgrade version of Maven compiler plugin

eyalg1972:
The old version of compiler plugin does not support well JDK 1.7, so I upgraded it to 3.1 that support.


ctrueden:
Thanks!

PR #260, opened 2017-01-22
Add support for Assembler code compilation in MSVC on WIN64

eyalg1972:
Since in Win 64 bits "inline" are no longer supported, dev needed to compile Assembler code instead.

Now THis was for now added only to Windows 64 bits and MSVC, if required- it can be extendded to more Platforms

### DIFF
--- a/src/main/java/com/github/maven_nar/AbstractNarMojo.java
+++ b/src/main/java/com/github/maven_nar/AbstractNarMojo.java
@@ -320,9 +320,9 @@ public abstract class AbstractNarMojo extends AbstractMojo implements NarConstan
 
     this.architecture = NarUtil.getArchitecture(this.architecture);
     this.os = NarUtil.getOS(this.os);
-    this.msvc.setMojo(this);
     this.linker = NarUtil.getLinker(this.linker, getLog()); // linker name set in NarUtil.getAOL if not configured
     this.aolId = NarUtil.getAOL(this.mavenProject, this.architecture, this.os, this.linker, this.aol, getLog());
+    this.msvc.setMojo(this);
 
     final Model model = this.mavenProject.getModel();
     final Properties properties = model.getProperties();


src/main/java/com/github/maven_nar/NarCompileMojo.java
```diff
@@ -177,6 +177,22 @@ private void createLibrary(final Project antProject, final Library library)
...
     }
 
+      if(getOS().equals( OS.WINDOWS ) && getArchitecture().equals("amd64"))
+      {
+
+          int noOfASMSources = getSourcesFor(getAssembler()).size();
+
+          if(noOfASMSources > 0)
+          {  // Assmbler files exist
+
```
GregDomjan: We have not included such a source file existance pre check on any of the other compiler types, not clear on need for this addition here.
With inclusion here getSourcesFor asm is happening twice, value is stored here but not re used.

src/main/java/com/github/maven_nar/NarCompileMojo.java
```diff
@@ -177,6 +177,22 @@ private void createLibrary(final Project antProject, final Library library)
       }
     }
 
+      if(getOS().equals( OS.WINDOWS ) && getArchitecture().equals("amd64"))
```
GregDomjan: getAssembler should return null if not available, this test for os seems unecessary. 
Maybe arch also unecessary - why only 64 bit? while 32 bit assembler might generally be inline could it not also be in seperate files?

src/main/java/com/github/maven_nar/cpptasks/compiler/AbstractProcessor.java
```diff
@@ -67,7 +67,16 @@ protected static String getIdentifier(final String[] command, final String fallb
 
   protected AbstractProcessor(final String[] sourceExtensions, final String[] headerExtensions) {
     this.sourceExtensions = sourceExtensions.clone();
-    this.headerExtensions = headerExtensions.clone();
+      if(headerExtensions != null)
```
GregDomjan: Rather than changing the contract that all other compilers match, perhaps follow existing pattern and pass 
`new String[] {}`

src/main/java/com/github/maven_nar/cpptasks/msvc/Assembler64bitCompiler.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public final class Assembler64bitCompiler extends CommandLineCompiler {
+    private static final Assembler64bitCompiler instance = new Assembler64bitCompiler(
+            false, null);
+    public static Assembler64bitCompiler getInstance() {
+        return instance;
+    }
+    private String identifier;
+    private Assembler64bitCompiler(boolean newEnvironment, Environment env) {
+        super("ml64", null, new String[]{".asm"}, null, ".obj", false, null, newEnvironment, env);
```
GregDomjan: Rather than null list of headers that requires change to Abstract class pass `new String[] {}`  ie. 
`super("ml64", null, new String[]{".asm"}, new String[] {}, ".obj", false, null, newEnvironment, env);`

eyalg1972:
The problem we have is the some modules needs the asm files, but only for
Windows 64 bits (And not for Linux, for example).

[ The motivation was the Microsoft in Win64 bit removed the support of
"inline", so our dev for Win64 only added Assembler]

This is why we put limitation on the OS and ARCH. if we disable them- we
need another way to verify that the

Eyal

src/main/java/com/github/maven_nar/cpptasks/msvc/Assembler64bitCompiler.java
eyalg1972: Sound OK, Who needs to modify that?

src/main/java/com/github/maven_nar/NarCompileMojo.java
eyalg1972: How can we limit that ASM files will be compiled only on Specific Platform? We need to have this ability since currently we are only Compiling them on Windows 64 bit, while other Platforms- WIndows 32 bit, Linux, AIX and Solaris- We don't need them.

src/main/java/com/github/maven_nar/NarCompileMojo.java
eyalg1972: Since it is rare cases where ASM files will exist, and compiled, and linked with other C/C++ objects - There is no need every time to add the relevant Assembler compiler, only if the actual files exist

eyalg1972:
I implemented Greg suggestions about AbstractProcessor.java

ctrueden:
Looks like this PR needs to be rebased against the latest master. @eyalg1972 Can you please rebase this over the latest master, squashing these two commits together, and then force push it?

```
git fetch
git checkout assembler_win64
git rebase -i master
```
In the editor, change the second `pick` to `fixup`, save and quit.

You should end up with one commit on top of master. Although you will probably get conflicts which must be fixed first. Edit the conflicting areas, `git add` the resulting files, then `git rebase --continue` when all conflicts are resolved.

Finally, do `git push --force` on the result, and that will update this PR.

Then hopefully @GregDomjan can rereview it.

eyalg1972:
Hi,

The problem here is that I use the master branch for all my commits, so
this rebase will create a different thing.
[ I need to have one branch with all my commits that I will actually use
for building]

Can I instead merge the changes from branch- AIX_g++_support (Which you
already approved) and then do rebase on it?

(I assume the existing conflicts are limited to
src/main/java/com/github/maven_nar/cpptasks/compiler/AbstractProcessor.java)

Eyal

eyalg1972:
I use feature branch for every commit I do, but I need one central branch
that all the commits needs to be made so I can work with one version.

I am not sure doing the rebase with the master will be helpfull, since on
my master there are commits you have not merged yet, I will try, but if it
will not work- the other option I mentioned should also resolve the
conflict, no?

Eyal

eyalg1972:
Hi

New PR was created-  Add support for Assembler code compilation in MSVC on
WIN64 #271

Eyal

PR #262, opened 2017-02-01
Add default define for Win AMD64 and few options to Linux

eyalg1972:
Add options that are mandatory for Linux and optional define for Windows


ctrueden:
Thank you!

PR #263, opened 2017-02-01
Add support for g++ on AIX

eyalg1972:
n AIX the objects needs to be created with ".a" extension, not ".so" like on Linux/Solaris. Also- needs to had support in the NAR it self to the configuration of AIX


ctrueden:
Looks great, thank you!

PR #269, opened 2017-04-23
Add support in compiling using VS 2005 with build-in SDK

eyalg1972:
Currently the NAR plugin does not support running using VS 2005 with it's default SDK, the purpose of the commit is to allow it.


ctrueden:
Thanks for your contribution, @eyalg1972!

It was a bit messy, but I cleaned it up. Merge commit is 436a5cc07ad6d54972e719e8f23d0265593632e2.

PR #272, opened 2017-05-04
Extend MSVC setup to work with VC 2017.

dmakarov:
The MSVC settings for compiler and linker currently don't work with Visual Studio 2017.  This PR extends the code that sets up the paths to the compiler, linker, include and library directories so that this works for VS 2017 configured for 64-bit compilation.

Would it be possible to make a new release of the plugin with these changes soon?

### DIFF
--- a/src/main/java/com/github/maven_nar/Msvc.java
+++ b/src/main/java/com/github/maven_nar/Msvc.java
@@ -132,6 +132,12 @@ public class Msvc {
       envVariable.setKey("TMP");
       envVariable.setValue(getTempPath());
       task.addEnv(envVariable);
+      final String envInclude = System.getenv("INCLUDE");
+      if (envInclude != null) {
+        for (final String path : envInclude.split(";")) {
+          addIncludePathToTask(task, new File(path));
+        }
+      }
     }
   }
 
@@ -164,6 +170,12 @@ public class Msvc {
         for (File sdkLib : sdkLibs)
           linker.addLibraryDirectory(sdkLib, sdkArch);
 
+      final String envLib = System.getenv("LIB");
+      if (envLib != null) {
+        for (final String path : envLib.split(";")) {
+          linker.addLibraryDirectory(new File(path));
+        }
+      }
     }
   }
 
@@ -230,10 +242,17 @@ public class Msvc {
       toolPathLinker = new File(this.home, "VC/bin/" + osArchitecture + "_" + mojoArchitecture).getAbsolutePath();
     }
     if (null == toolPathLinker) {
-      if ("amd64".equals(mojoArchitecture))
+      if ("amd64".equals(mojoArchitecture)) {
         toolPathLinker = new File(this.home, "VC/bin/amd64").getAbsolutePath();
-      else
+        if (!new File(toolPathLinker).exists()) {
+          final String envVCToolsInstallDir = System.getenv("VCToolsInstallDir");
+          if (envVCToolsInstallDir != null) {
+            toolPathLinker = new File(envVCToolsInstallDir, "bin/HostX64/x64").getAbsolutePath();
+          }
+        }
+      } else {
         toolPathLinker = new File(this.home, "VC/bin").getAbsolutePath();
+      }
     }
     if ("amd64".equals(osArchitecture) && !matchMojo) {
       addPath(this.home, "VC/bin/amd64");
@@ -444,6 +463,7 @@ public class Msvc {
       File includeDir = new File(kitDirectory, "Include/" + version);
       File libDir = new File(kitDirectory, "Lib/" + version);
       addSDKLibs(includeDir, libDir);
+      setKit(kitDirectory);
     }
   }
 


ctrueden:
Thanks!

ctrueden:
> Would it be possible to make a new release of the plugin with these changes soon?

I will try. The release infrastructure I was using to make releases of NAR is currently broken. (For the curious: I have a Jenkins job to do it; we recently upgraded our Jenkins due to several high-impact CVEs; and now the job is broken.) I am working on establishing a new system for doing releases, which will allow me to make a new release of NAR. But it may be several days yet.

dmakarov:
Thank you!

GregDomjan:
While it's a step forward, this change doesn't completely fix for VS 2017 - cross compile and x86 are not working. 
Also makes use of environment values to work around not finding other information.

Earlier versions added the VS<ver>COMNTOOLS values to global, 
> VS100COMNTOOLS=c:\Program Files (x86)\Microsoft Visual Studio 10.0\Common7\Tools\
> VS110COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio 11.0\Common7\Tools\
> VS120COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio 12.0\Common7\Tools\
> VS140COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\
> VS80COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio 8\Common7\Tools\

for VS 2017 they are only added with the developer command prompt or running the vcvarsall.bat

> VS150COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\
> VCToolsInstallDir=C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.10.25017\

looking up where VS and VC needs something that doesn't require launching maven inside a batch file so vcvarsall can be called first.


Previously this.home -> VS120COMNTOOLS . parent.parent  would have bin/include/lib folders directly under the VC folder, for VS 2017 there is new structure Tools\MSVC\14.10.25017\

Further the layout of bin folders has changed to 
have Host<arch>\<arch>  
instead of <arch> or <arch>_<arch>

old x86_x64  - new hostx86\x64
old amd64   - new host hostx64\x64
old amd64_arm - new hostx64\arm

I'll be trying to address some of these shortly

src/main/java/com/github/maven_nar/Msvc.java
```diff
@@ -132,6 +132,12 @@ public void configureCCTask(final CCTask task) throws MojoExecutionException {
       envVariable.setKey("TMP");
       envVariable.setValue(getTempPath());
       task.addEnv(envVariable);
+      final String envInclude = System.getenv("INCLUDE");
```
dmakarov: According to Microsoft documentation https://docs.microsoft.com/en-us/cpp/build/reference/cl-environment-variables CL compiler uses the environment variable `INCLUDE`, and if it's set incorrectly that already adds uncertainty.  So, it may be not the plugin's responsibility to work around messy environment settings.  It might be redundant to add paths from `INCLUDE` variable on the command line as `-I` options, except that `cl` as run by maven is executed in an environment which does not contain `INCLUDE` in it.

src/main/java/com/github/maven_nar/Msvc.java
```diff
@@ -164,6 +170,12 @@ public void configureLinker(final LinkerDef linker) throws MojoExecutionExceptio
         for (File sdkLib : sdkLibs)
           linker.addLibraryDirectory(sdkLib, sdkArch);
 
+      final String envLib = System.getenv("LIB");
```
dmakarov: `LIBPATH` might be a better choice, indeed, unless you want to ignore all environment variables.

GregDomjan: As a more concrete example, here is an excerpt from my log trying to use this
Trying to build x86+x64 package, I happened to run it in a shell that was opened with VCx64 dev setup which worked fine previously, but now fails.
The additional libs in the search path result in incorrect libs being chosen and fully pathed

> [DEBUG]   (f) architecture = x86
> 
> '/MACHINE:X86'
> '/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\lib'
> '/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\atlmfc\lib'
> '/LIBPATH:C:\Program Files (x86)\Windows Kits\8.1\Lib\winv6.3\um\x86'
> '/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\lib\amd64'
> '/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\atlmfc\lib\amd64'
> '/LIBPATH:C:\Program Files (x86)\Windows Kits\8.1\Lib\winv6.3\um\x64'
> '/NOLOGO'
> 
> '"C:\Program Files (x86)\Windows Kits\8.1\Lib\winv6.3\um\x64\Secur32.Lib"'
> '"C:\Program Files (x86)\Windows Kits\8.1\Lib\winv6.3\um\x64\Crypt32.Lib"'
> 
>  error LNK2001: unresolved external symbol __imp__CryptProtectData@28
>  error LNK2001: unresolved external symbol __imp__CryptUnprotectData@28

Adding the LIBPATH or LIB environment is particularly poor for building multi targeted builds.

src/main/java/com/github/maven_nar/Msvc.java
GregDomjan: The doc does indeed say that `CL` will use `INCLUDE` and that it should be set to the VC include directory.

Experimentation so far is showing that `INCLUDE` doesn't really need to be set.

The change you have made though isn't setting the `INCLUDE` environment for the `CL` execution, rather it is taking the existing environment and adding it to the `CL` command line, thus overriding part of the point of discarding the environment to run the tools (more) cleanly.

I don't wish to just break whatever you got working, but I could use more of an idea of what includes where actually missing for you.

src/main/java/com/github/maven_nar/Msvc.java
dmakarov: Perhaps then, the plugin should check whether the variables are defined in the environment in which `mvn` command is executed and pass these variables into the environment in which `cl` or `link` of cpp or link task are executed.  Currently, these variables are ignored and paths to header files and libraries are (somewhat arbitrarily) guessed and passed as command line arguments in the corresponding commands.  The assumption is that if `INCLUDE` or `LIBPATH` are set, they are set by running one of the VS `vars` scripts and define an environment for specific kind of build, be it a 32 or 64-bit target or cross-compilation build.

GregDomjan: This solution of hoisting in any global setting and not recording the configuration leads to variations in the build across build environments, which is undesirable and to be avoided where possible.
Scenarios of note:
 * build multi arch (x86/x64/arm) from the same command line
 * build multi linker (VC8/9/10/11/12/14/15) from the same command line
 * build server builds where you cannot (or it's difficult to) change the build environment before calling mvn, and where one environment might not be enough
 * vagaries between build hosts leading to incorrect headers/libs being included.

I've started a branch for VS 2017 changes, it was able to build 32 and 64 regardless of env
https://github.com/maven-nar/nar-maven-plugin/tree/feature/MSVC-VS2017

Includes refactoring and finding the appropriate target architecture specific libs as well as app paths.
I will be testing the plugin on a larger build as I resolve other codebase upgrade issues to VC15 with it, see if any other issues come out.  
Would be happy to hear any inputs on parts still needing to be resolved, or if something would then be broken for your build.

PR #273, opened 2017-05-11
Add support in pure 32 bits compilation on Windows 64 bits machine

eyalg1972:
Currently if We compile on Windows 64 bits machine a 32 bit application, the NAR choose to use the amd64_x86 configuration.
(Same- if We compile on 32 bits machine 64 bits it will choose x86_amd64)

Now a new flag could be used to choose just regular 32 bits (Or 64 bits) without the mixing between them


ctrueden:
Thanks @eyalg1972!

eyalg1972:
You Welcome!!

PR #274, opened 2017-05-11
Add support for HPE Fortify code scanning during compilation

eyalg1972:
We need to use HPE Fortify to scan the C/C++ code, this scanning is done during compilation by adding prefix - sourceanalyzer -b <FortifyID string>

The way I implemented it was to use the new attribute existence as a flag that we need to use the Fortify prefix

### DIFF
--- a/src/main/java/com/github/maven_nar/AbstractCompileMojo.java
+++ b/src/main/java/com/github/maven_nar/AbstractCompileMojo.java
@@ -121,6 +121,16 @@ public abstract class AbstractCompileMojo extends AbstractDependencyMojo {
   @Parameter
   private Java java;
 
+    /*
+  * To support scanning the code with HPE Fortify
+  * The attribute is used both as a flag that Fortify is required and the value set is used for the
+  * When setting a value -  sourceanalyzer –b <fortifyID> will be prepended to the
+  * command line
+  * */
+  @Parameter(defaultValue = "")
+  private String fortifyID;
+
+  
   /**
    * Flag to cpptasks to indicate whether linker options should be decorated or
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/maven_nar/NarCompileMojo.java
```diff
@@ -179,6 +179,8 @@ private void createLibrary(final Project antProject, final Library library)
     if (getCpp() != null) {
       final CompilerDef cpp = getCpp().getCompiler(Compiler.MAIN, null);
       if (cpp != null) {
+		// Set FortifyID attribute
```
ctrueden: Please watch the whitespace. It does not match.

src/main/java/com/github/maven_nar/AbstractCompileMojo.java
```diff
@@ -121,6 +121,16 @@
   @Parameter
   private Java java;
 
+    /*
```
ctrueden: Note that this is not javadoc, but should be. And note that javadoc is HTML, so line breaks mean nothing without `<p>` or `<br>` tags. Best is to use the IDE to preview the final result.

ctrueden:
Thanks! I'm merging as is, and I'll do some minor cleanup on top of it.

In the future, please check the diff of each commit very carefully. The better and cleaner they look, the faster we can merge things, for several reasons.

ctrueden:
Cleaned up in a7cfe6157ebed7aed6cbadc1dcb563106b0a12a9 and 9884e30785fb348bda047808b31c12694bd79952.

PR #276, opened 2017-05-15
Find first the exact SDK request and add the forcing of 32 bit build to VS 2005

eyalg1972:
Currently when I specify to use version 6.0A it search in the loop on-

C:/Program Files (x86)/Windows Kits
C:/Program Files (x86)/Microsoft SDKs/Windows
C:/Program Files/Windows Kits
C:/Program Files/Microsoft SDKs/Windows

Now, since it ignores only higher versions of SDK (Not Lower), It detect version 5.0 under 
C:\Program Files (x86)\Microsoft SDKs\Windows

Before it detects-
C:\Program Files\Microsoft SDKs\Windows\v6.0A
This means it use version 5.0 when I asked for 6.0A.

The solution is to do two loops instead of one-
1) Serach for exact SDK version
2) If no SDK was found in the first loop- Try to find lower version

I am not sure why we need lower version of SDK, but this is how it was implemented first, so I did not want to cancel this option.

### DIFF
--- a/src/main/java/com/github/maven_nar/Msvc.java
+++ b/src/main/java/com/github/maven_nar/Msvc.java
@@ -115,10 +115,12 @@ public class Msvc {
   }
 
   static boolean isMSVC(final String name) {
+	
     return "msvc".equalsIgnoreCase(name);
   }
 
   public void configureCCTask(final CCTask task) throws MojoExecutionException {
+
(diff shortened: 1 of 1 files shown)


ctrueden:
Thanks @eyalg1972!

PR #278, opened 2017-05-21
Support working with non default linkers

eyalg1972:
Currently when the aol.properties contains a definition of linker for specific OS+ARCH (For example- ppc64.AIX.linker=g++) , NAR is by default trying to use that linker and ignore the linker that is configured (If there is a different configuration).

The fix is to first check if a linker was defined, and only if not- choose the default configuration


ctrueden:
Thanks!

PR #279, opened 2017-05-21
Add support to xlC_r compiler on AIX

eyalg1972:
Add support for IBM xlC_r compiler for AIX

### DIFF
--- a/src/main/resources/com/github/maven_nar/aol.properties
+++ b/src/main/resources/com/github/maven_nar/aol.properties
@@ -127,9 +127,7 @@ amd64.Windows.msvc.res.excludes=
 
 amd64.Windows.msvc.idl.compiler=midl
 amd64.Windows.msvc.idl.defines=
-amd64.Windows.msvc.idl.options=
 amd64.Windows.msvc.idl.includes=**/*.idl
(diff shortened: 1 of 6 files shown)


src/main/resources/com/github/maven_nar/aol.properties
```diff
@@ -127,9 +127,7 @@ amd64.Windows.msvc.res.excludes=
 
 amd64.Windows.msvc.idl.compiler=midl
 amd64.Windows.msvc.idl.defines=
-amd64.Windows.msvc.idl.options=
```
ctrueden: Are these removals intended? If so, I think the commit message needs to explain what this is relevant to the addition of xlC_r. Same for the removal of the `x86.Windows.gpp` lines.

eyalg1972:
I just cleaned some duplication.... (You can still find these lines, now just in one place instead of 2)

ctrueden:
> I just cleaned some duplication

Ah, I see, OK. Great. If you can do such things as separate commits, that would be ideal, but in the interest of pragmatism I'll merge as-is.

Thank you!

eyalg1972:
Next time I promise a separate  commit and PR....

Eyal

Issue #280, opened 2017-05-22
Enhancement - The nar plugin should provide an option to restrict dependencies to direct dependencies only


aciaccio:
It is desirable for a project to specify strictly those dependencies it needs and none other. Consider the following case:

Project A depends on static library B and shared library C. 

Shared library C also depends on static library B.

Project A can invoke APIs from B (directly, not transitively via C) and C because NAR accumulates every dependency (transitive and direct) at compile/link time even if Project A does not specify B as one of it's direct dependencies (i.e. in Project A's pom.xml). In this way if C was ever refactored to NOT depend on B then downstream projects (Project A for instance) will now fail to build and any audits to determine B's consumption would be obscured.

To help force a more constrained dependency strategy (i.e. being complete about your dependencies) a flag should be added to restrict a project's dependencies to only those that the project depends on. 

For instance, in the example above Project A must specify BOTH Project B and Project C as dependencies. If, for example, Project A specified C but NOT B as a dependency then B would not be included in the compilation/linking phase and should cause Project A's build to fail.

sdedwards:
There is no need for this feature in nar-maven-plugin because Maven itself provides everything you need to control your dependencies. Take a look at how to use  the ["optional" and "excludes" dependencies](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) features.

In your example, the pom.xml of shared library C should declare the dependency on static library B as "optional" because the API of B should not be exposed to users of C. Doing this will prevent Maven including B in project A's dependency list by default unless of course project A wants to include B explicitly. This is how you should manage internal components of shared libraries.

Dependency "exclusions" can be used as a last resort if you have no control over Project C. In Project A you can simply exclude B from the dependency tree and NAR will ignore it.

GregDomjan:
It seems like your trying to force a method of resolving poor documentation hygiene through breaking build.  

Make it a bit more complexe/complete
Project A & shared library C both depend on static library B and shared library D. 
A also depends on C.  
D also depends on other things like static E
B also depends on other things like static F

To compile A you need A,C,B,D,F
To test A you need A,C,D to run

if C needs a newer B, then depending on usage maybe A should change or not.
if C needs a newer D, A should probably use the same... 
  with ref specified in A and C to D means you have to resolve conflict
  with ref only specified in C to D means you have potential instability in A if C changes with identified conflict... 

[...]

PR #270, opened 2017-04-28
Committing changes for dependency library ordering enhancement

aciaccio:
The changes here resolve issue 265:
https://github.com/maven-nar/nar-maven-plugin/issues/265

All new file additions are integration tests that validate the proposed changes. This enhancement adds a new dependency on Aether (see the pom.xml in the parent directory).  

Any and all feedback is welcome.

PR #283, opened 2017-06-16
Adding support for forcing direct dependencies

aciaccio:
The changes here fix issue 280:
Enhancement - The nar plugin should provide an option to restrict dependencies to direct dependencies only #280

Included is a set of automated tests that verify the new enhancement. Any and all feedback is welcome. Thanks folks!

### DIFF
--- a/src/it/it0032-directDepsOnly-fail/invoker.properties
+++ b/src/it/it0032-directDepsOnly-fail/invoker.properties
@@ -1,0 +1,2 @@
+# The expected result of the build for this project we WANT the project to fail.
+invoker.buildResult = failure
--- a/src/it/it0032-directDepsOnly-fail/pom.xml
+++ b/src/it/it0032-directDepsOnly-fail/pom.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  #%L
+  Native ARchive plugin for Maven
+  %%
+  Copyright (C) 2002 - 2014 NAR Maven Plugin developers.
+  %%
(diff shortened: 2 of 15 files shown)


ctrueden:
I am traveling now but will try to find time to review this before end of June. In the meantime, if anyone else has time to do a review, that would be much appreciated!

ctrueden:
@GregDomjan @sdedwards Since you spoke up on the issue #280: do you have any concerns or reservations about merging this? Ultimately I don't have a good intuition for whether adding this is a Bad Idea™. From my perspective, as long as backwards compatibility of existing build configuration and public API is maintained, and there are integration tests exercising it, merging seems OK to me, since it addresses @aciaccio's requirements.

sdedwards:
@ctrueden I suppose giving the users more options to control dependencies isn't a Bad Idea. It definitely should not be the default behaviour though. Note that Maven will still try to resolve the whole dependency tree even with the `directDepsOnly` option set, i.e. the local repository must have all the libraries available even though we are only compiling/linking to the direct ones.

ctrueden:
Thanks for the feedback, @sdedwards. I'll merge now so that @aciaccio can get this feature without further delays.

aciaccio:
+1 to *not* being the default behavior. Yes Maven will still resolve the
dependency tree (although I believe using the fixes from
https://github.com/maven-nar/nar-maven-plugin/pull/270 are ultimately
required to assure us the link line is properly ordered). The changes
included with directDepsOnly happen at link line creation time, not
dependency resolution time. directDepsOnly simply enforces constraints when
creating the link line.

Issue #291, opened 2017-10-06
Unable to pass arguments containing spaces to gnuConfigureArgs for nar-gnu-configure goal.


sraza1:
There can be a use case where some of the configurations arguments passed to gnu configure script might contain white spaces in them. An example can be project names or other string based properties. e.g 
--with-project-name= My Project
 
The <gnuConfigureArgs> seems to be treating the spaces as separate arguments even when the arguments are enclosed between quotes i.e "--with-project=My Project".

sraza1:
Submitted PR : https://github.com/maven-nar/nar-maven-plugin/pull/292

PR #292, opened 2017-10-06
* Allow gnuConfigureArgs to accept configure arguments with spaces.

sraza1:
Currently, the arguments are split using space as a delimiter. With this change arguments
  containing spaces but specified within quotes are treated as a single argument.

### DIFF
--- a/src/main/java/com/github/maven_nar/NarGnuConfigureMojo.java
+++ b/src/main/java/com/github/maven_nar/NarGnuConfigureMojo.java
@@ -21,6 +21,9 @@ package com.github.maven_nar;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.regex.*;
 
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
@@ -154,7 +157,14 @@ public class NarGnuConfigureMojo extends AbstractGnuMojo {
 
         // create the array to hold constant and additional args
         if (this.gnuConfigureArgs != null) {
-          final String[] a = this.gnuConfigureArgs.split(" ");
+          List<String> list = new ArrayList<String>();
+          Matcher m = Pattern.compile("([^\"]\\S*|\".+?\")\\s*").matcher(this.gnuConfigureArgs);
+          // strip double quotes
+          while (m.find()) {
+            list.add(m.group(1).replace("\"", ""));
+          }
+          String[] a = new String[list.size()];
+          a = list.toArray(a);
           args = new String[a.length + 2];
 
           System.arraycopy(a, 0, args, 2, a.length);


ctrueden:
Thanks for submitting this fix, @sraza1!

Would you have time to create or modify one of the [integration tests](https://github.com/maven-nar/nar-maven-plugin/wiki/Testing#running-the-integration-tests) to illustrate this issue? The goal is to have a test which fails prior to your patch, but passes after it.

sraza1:
> Thanks for submitting this fix, @sraza1!
>Would you have time to create or modify one of the integration tests to illustrate this issue? The goal is to have a test which fails prior to your patch, but passes after it.

Sure I'll get back with a test case shortly ...

Thanks,
Saqlain

src/main/java/com/github/maven_nar/NarGnuConfigureMojo.java
```diff
@@ -154,7 +157,14 @@ public final void narExecute() throws MojoExecutionException, MojoFailureExcepti
 
         // create the array to hold constant and additional args
         if (this.gnuConfigureArgs != null) {
-          final String[] a = this.gnuConfigureArgs.split(" ");
+          List<String> list = new ArrayList<String>();
+          Matcher m = Pattern.compile("([^\"]\\S*|\".+?\")\\s*").matcher(this.gnuConfigureArgs);
+          // strip double quotes
+          while (m.find()) {
+            list.add(m.group(1).replace("\"", ""));
```
GregDomjan: This seems a bit imprecise removing any "
Maybe you wouldn't use a " in gnuconfigure arg, or escape one in an arg?

sraza1: >Maybe you wouldn't use a " in gnuconfigure arg, or escape one in an arg?
So, keeping in view https://github.com/maven-nar/nar-maven-plugin/issues/291 would searching for the starting " and ending " and stripping them only (if found) be the better way of doing this ?

ctrueden:
@sraza1 Are you still interested in submitting a test case to validate this bug-fix? It would be very much appreciated.

sraza1:
@crtueden Yes, I would surely like to do that. Hopefully I'll be able to put up something during the next week...

sraza1:
@ctrueden I have updated the PR for an integration test. Kindly take a look.

Thanks,
Saqlain

src/main/java/com/github/maven_nar/NarGnuConfigureMojo.java
GregDomjan: Sorry, totally forgot to come back to this. 
Maybe I misread the regex, or it changed?

Could the args include an escaped `\"` quote?  does it even need to be handled, I don't know.

taking the unit test and expanding with some other junk 
```
<gnuConfigureArgs>
		"--with-pkginfo=GNU Executable Integration Test"  arg2 "arg 3"
"arg\" 4"
	</gnuConfigureArgs>
```

Expected 4 args
 * --with-pkginfo=GNU Executable Integration Test
 * arg2
 * arg 3
 * arg\" 4

Regex provided will come back with 6 matches 
 1 that's empty spaces/newlines
 2-4 that need the leading/trailing " removed
 5/6 should actually just be 5 and needing " removed.

I don't do much regex in Java, but I believe from doc this should work, and the capture group should already exclude the "
`\s*\"(.*?)(?<!\\)\"\s*`  or as Java string `"\\s*\\"(.*?)(?<!\\\\)\\"\\s*`
Checked using https://regex101.com/

PR #293, opened 2017-10-16
Feature/msvc vs2017

GregDomjan:
Improved Visual Studio 2017 / VC & Windows 10 SDK support


GregDomjan:
@dmakarov  would you mind checking if this satisfies what you where aiming for as well?

PR #295, opened 2017-11-16
Transitive deps

garlockchristopher:
This change will provide the linker with the path to all of the shared objects that are transitive dependencies if directDepsOnly is enabled.


garlockchristopher:
Didn't realize this was going to create a pull request to the forked repo, closing it till I have completed some more testing.

PR #298, opened 2017-12-04
Provide Linux linker with path to transitive dependencies when directDepsOnly is true.

garlockchristopher:
The direct deps only flag gives us control over which libraries get added to the link line, but was also forcing us to specify more dependencies than were really needed.  With these changes we will be able to detect when a transitive dependency will be needed by the linker, and pass the path to the shared library to ld so it will know where to look for the symbols.  It also reduces memory utilization when parsing through the full dependency list by removing duplicates early on in the process.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
   <groupId>com.github.maven-nar</groupId>
   <artifactId>nar-maven-plugin</artifactId>
-  <version>3.6.4-SNAPSHOT</version>
+  <version>3.7.0-SNAPSHOT</version>
   <packaging>maven-plugin</packaging>
 
   <name>Native ARchive plugin for Maven</name>
--- a/src/it/it0033-directDepsOnly-link-transitive-deps/it0033-directDepsOnly-link-transitive-deps-direct/pom.xml
+++ b/src/it/it0033-directDepsOnly-link-transitive-deps/it0033-directDepsOnly-link-transitive-deps-direct/pom.xml
@@ -1,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  #%L
+  Native ARchive plugin for Maven
+  %%
+  Copyright (C) 2002 - 2014 NAR Maven Plugin developers.
+  %%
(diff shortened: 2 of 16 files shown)


ctrueden:
Thanks for the patch, and sorry for the long delay in reply. I just got back from vacation, and before that a hackathon, so I am way behind on email etc. I fear I will not have time to review this within the next several weeks. In the meantime, review & feedback from other contributors/maintainers is welcome.

AnnDinKC:
@ctrueden Could we get an updated estimate from you on when you think you'll be able to take a look at this pull request? Thanks!

AnnDinKC:
@ctrueden  bump

ctrueden:
I hear your bumps, @AnnDinKC. I will try to go through the NAR backlog next
week. I have over 300 read-but-pending-action emails sitting in my inbox --
the most backlogged I have ever been. Is it just my imagination, or is the
pace of technology getting even faster?

ctrueden:
There is a lot of noise in the reported commits above. I'd like to see this PR rebased against the latest master. @garlockchristopher Could you do `git fetch && git checkout direct-deps-link-transitive-deps && git rebase origin/master && git push --force` please?

garlockchristopher:
@ctrueden Okay It looks like the conflicts have been resolved.

ctrueden:
@garlockchristopher Great. I'm still confused about the merge commits, though. I'm pulling your branch locally to try to get a feel for what happened here.

ctrueden:
OK, I rebased it over master. There are still a few commits to go through though; will review shortly.

ctrueden:
There are a lot of small things I would change, like squashing the commits that introduce the `trimDuplicates` function and then remove it again in favor of the `removeAll` call, and using Java 7's diamond syntax, and making more things into separate methods (such as the check for whether to reverse the dependencies list).

But none of those things are showstoppers. I think this branch can be merged as is, and if the code style bugs people later, they can clean it up as they wish.

So thank you @garlockchristopher for your work on this! Since several topic branches just merged: please test your projects now with the latest post-merge master branch, and report back if there are any issues.

PR #299, opened 2018-01-03
integration test showing working and not working case for secondary executions

AnnDinKC:
There is a bug in secondary executions where, when using the <options combine.children="append"> tag in the cpp block of a secondary execution block, the default execution cpp options are lost to the secondary execution.


ctrueden:
So sorry for the long delay. And thank you very much, @AnnDinKC!

PR #303, opened 2018-02-23
Support deployment of artifacts from multiple platforms/AOLs.

mangrer:
Please refer issue, https://github.com/maven-nar/nar-maven-plugin/issues/304


ctrueden:
Thank you @mangrer, for the thorough writeup of the problem, explanation of solution, and integration tests! Very much appreciated!

PR #305, opened 2018-02-27
Add environment variables to tests

### LINKED ISSUE #281: Enhancement - Pass environment variables to unit tests
In some cases it is necessary to set environment variables when running a test but it seems that this is currently not supported by the nar plugin. The Maven surefire plugin support this already using the &lt;environmentVariables&gt; tag, but it cannot be used to run C++ unit tests.

Specifically, I have an external library which dynamically loads custom plugins from a location that can be specified using an environment variable. I can't install the plugins to the default plugin location as part of the Maven build since it is under a system folder that requires root privileges to write. The only possibility that I see at the moment (other than changing the unit test binary to pass the variable as a command line argument and set it in the code) is to execute the test binary within a shell script (called via `exec-maven-plugin`), but this is not a portable solution.

Other use cases would be to specify a different locale, time zone, etc. when running the tests.

Zetten:
Allow NAR tests to define a map of environment variables to be set
during execution.

Fixes #281.

### DIFF
--- a/src/it/it0007-lib-shared/pom.xml
+++ b/src/it/it0007-lib-shared/pom.xml
@@ -60,6 +60,9 @@
             <test>
               <name>HelloWorldTest</name>
               <link>shared</link>
+              <environmentVariables>
+                <TESTENVVAR>test</TESTENVVAR>
+              </environmentVariables>
             </test>
           </tests>
         </configuration>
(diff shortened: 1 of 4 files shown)


ctrueden:
@Zetten Thank you very much!

Issue #309, opened 2018-03-22
The TestLinkerVersion unit test fails on AIX.


garlockchristopher:
This test is failing because it tries to get the version of the gcc linker.  
`testVersion(com.github.maven_nar.test.TestLinkerVersion)  Time elapsed: 0.324 sec  <<< ERROR!
org.apache.maven.plugin.MojoExecutionException: Could not launch /bin/sh -c 'gcc' '--version'
        at java.lang.UNIXProcess.<init>(UNIXProcess.java:291)
        at java.lang.ProcessImpl.start(ProcessImpl.java:167)
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1040)
        at java.lang.Runtime.exec(Runtime.java:631)
        at org.codehaus.plexus.util.cli.Commandline.execute(Commandline.java:655)
        at com.github.maven_nar.NarUtil.runCommand(NarUtil.java:590)
        at com.github.maven_nar.NarUtil.runCommand(NarUtil.java:556)
        at com.github.maven_nar.Linker.getVersion(Linker.java:569)
        at com.github.maven_nar.Linker.getVersion(Linker.java:549)
        at com.github.maven_nar.test.TestLinkerVersion.testVersion(TestLinkerVersion.java:56)
`
The xlc linker is invoked when building projects with the nar-maven-plugin on aix, so the issue seems to be isolated to the unit test.

Issue #308, opened 2018-03-22
The directDepsOnly property does not apply to test scope dependencies.


garlockchristopher:
If you specify directDepsOnly in the pom file, transitive dependencies of dependencies with `<scope>test</scope>` do not get trimmed.

PR #310, opened 2018-04-06
Apply directDepsOnly changes to test scope

garlockchristopher:
This pull request addresses issue #308 and provides a work around for issue #309 so we can build on AIX without the failing unit test blocking the build.


ctrueden:
@garlockchristopher Really sorry this sat unanswered for so long!

Great patch. The ITs 0036 and 0037 are passing on Linux, so I'm merging this. (There are known issues with the ITs already on macOS; see #267.) :beers:

PR #311, opened 2018-04-13
Add 'sharedObjectName' flag to enable shared archives on AIX

garlockchristopher:
This enhancement will allow users to build shared archives out of their shared library projects. By setting the sharedObjectName property users will set the name of the shared object file which will be added to a shared archive using the output file naming rules before being added to nar file. I also added a test and documentation for this flag as well as two more flags which were recently added but not documented.

### DIFF
--- a/src/it/it-parent/pom.xml
+++ b/src/it/it-parent/pom.xml
@@ -62,4 +62,31 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  
+  <profiles>
+    <profile>
+      <id>aix-it-profile</id>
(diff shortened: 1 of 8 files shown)


ctrueden:
Thanks @garlockchristopher. And again sorry for the delay.

The commit message and author were strange, so I rewrote it with you as author, and a commit message based on your narrative from this PR. Then I cherry-picked it to master as ffe006d21c0c6bdc8105f10d4104e702e448e0c8. Therefore, the content of this PR is now merged! :beers:

PR #316, opened 2018-06-02
Add additionalParameters to javah, in order to use 'javac -h' on JDK10

eolivelli:


eolivelli:
This does not work.
javac -h is NOT like javah
because it works on .java files, non on compiled classed

eolivelli:
Actually it is enough to add an option to skip javah phase.
Maven compiler plugin can do javah work

PR #320, opened 2018-07-16
Removed redundant log entry that pollutes the build log

bchiodo:
I feel that the log statement is inappropriate because it is printed at INFO level so pollutes the build log and detracts from other useful information and/or errors. Even without this log statement the "executed command" can already been seen by passing the -X option.

### DIFF
--- a/src/main/java/com/github/maven_nar/cpptasks/CommandExecution.java
+++ b/src/main/java/com/github/maven_nar/cpptasks/CommandExecution.java
@@ -66,7 +66,7 @@ public class CommandExecution {
                 //Append space
                 builder.append(" ");
             }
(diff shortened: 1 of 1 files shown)


GregDomjan:
Not sure on this one, comment seems fair,  but would rather delete than just comment out.
Was thinking that change to debug rather than just removal would be better, but as -X already shows then it would just show twice in debug,  is it actually the same log.

zosrothko:
@bchiodo give a try to the latest nar-maven-plugin:3.9.3-SNAPSHOT where the <commandLogLevel>element is honored. Set it to 0 will produce a more quite log.

bchiodo:
@zosrothko This is verified working in v3.9.3+. Closing.

PR #325, opened 2018-08-20
added new option to skip running the ranlib program on attached archi…

bchiodo:
…ves when it would otherwise be run

### DIFF
--- a/src/main/java/com/github/maven_nar/AbstractNarLayout.java
+++ b/src/main/java/com/github/maven_nar/AbstractNarLayout.java
@@ -102,7 +102,7 @@ public abstract class AbstractNarLayout implements NarLayout, NarConstants {
   }
 
   protected void unpackNarAndProcess(final ArchiverManager archiverManager, final File file, final File narLocation,
-      final String os, final String linkerName, final AOL defaultAOL)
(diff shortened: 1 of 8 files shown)


src/main/java/com/github/maven_nar/AbstractNarLayout.java
```diff
@@ -133,10 +133,14 @@ protected void unpackNarAndProcess(final ArchiverManager archiverManager, final
...
     }
     if (linkerName.equals(gcc) || linkerName.equals(gpp)) {
-      NarUtil.runRanlib(new File(narLocation, "lib/" + defaultAOL), this.log);
+      if (!skipRanlib) {
+        NarUtil.runRanlib(new File(narLocation, "lib/" + defaultAOL), this.log);
+  	  }
       // FIXME clumsy
       if (defaultAOL.hasLinker(gpp)) {
-        NarUtil.runRanlib(new File(narLocation, "lib/" + NarUtil.replace(gpp, gcc, defaultAOL.toString())), this.log);
+        if (!skipRanlib) {
```
GregDomjan: Wonder if it's style, but why not just surround the whole rather than test !skipRanlib twice?

GregDomjan:
This seems ok to me asides from one style note.
As there are no rejections, and it passes travis, going with it.

PR #345, opened 2019-08-23
Issue 343 tool path windows sdk alainco

AlainCo:
merge nar version


AlainCo:
It was a mistake with github, Sorry.

I have written the proposed correction in issue 343, few lines to add. I imagine it is better to do it yourself than sync with my forked repo...

Issue #346, opened 2019-10-02
Unable to generate shared library from parent project with 8 threads


duckladydinh:
Hallo,

i have a big maven project and a small module called native which uses maven-nar-plugin. When I do something like mvn clean install -T 8, the nar-compile failed since it could not find some dependent files. But if I do use 4 threads or less, or do clean install directly in the native folder, it would be perfectly fine.

Does anyone experience the same issue? I would appreciate if you can give me a hint.

Thank you.

GregDomjan:
We build a project of about 150 modules, but only single-threaded due to issues like this.

The related problems seem to be
 * identifying the plugin as safe/unsafe for threaded usage
 * incomplete support of listing dependencies to the maven resolver - only the metadata jar is added
 * only tracking packed/installed artifacts - requiring install and subsequent unpack of dependency to properly refer to it

In theory, Maven can schedule a compile if all other dependencies have passed compile stage, but at that point, the artifact references in the reactor don't list the AOL dependencies. Also when they get added during download step when not already added in the reactor, they aren't looking for unpacked source dirs, only the .nar to unpack.

PR #349, opened 2019-11-27
A multi-threading problem and proposed fix

duckladydinh:
Hi, I would like to propose this change to fix the issue mentioned in #346. A better way may be to disable multi-threading for NAR projects, but I am not sure about it.


duckladydinh:
Hi, even though I fail to find a simple project that can be reproducible to upload to GitHub. This is indeed an issue that only happens in large-sized projects (It took me a total of 2 weeks just to reproduce, excluding debugging and fixing). Could someone give it a review?

At least, if it does not cause harm, I think this fix would be beneficial for many when the problem really occurs.

ctrueden:
Thanks @duckladydinh for the patch, and thanks @GregDomjan for looking at it.

I don't have much time these days to help maintain NAR, but I don't want to block progress. Merging as is. Other maintainers please feel welcome in future to merge things without me.

PR #328, opened 2018-09-18
Support Linux on PPC64LE

ChristophBerg:
Compiling https://github.com/tada/pljava on Debian's ppc64el
architecture failed because nar doesn't support the platform. Fix by
copying the amd64.Linux definitions and substituting s/amd64/ppc64le/g.

PR #354, opened 2020-03-27
Support aarch64 (arm64) Linux

df7cb:
Like in 086f6b3e (#328), this is to support compiling
https://github.com/tada/pljava, this time on the aarch64 (called arm64
in Debian) platform. This is the amd64.Linux definitios copied, the
architecture name replaced, and "-m64" removed because it's not a valid
flag on that architecture.


entropaos:
This patch only exists in the master branch, the latest release still doesn't support aarch64.
How can we use nar-maven-plugin to compile code for aarch64? Thank you, guys.

PR #355, opened 2020-04-19
Support Linux gcc on sparc64 and sparc

turelto:
Compiling https://github.com/tada/pljava on Debian's sparc64 and sparc
architectures failed because nar doesn't support those platforms. With
these properties pljava builds successfully on Debian sparc64 (tested on
Buster) and Debian sparc (tested on Jessie).


turelto:
Similar to 88237dd (#354) and 086f6b3 (#328), this patch supports compiling https://github.com/tada/pljava on sparc64 and sparc Linux platforms. The properties are based on amd64/i386 and were tested on Debian Buster sparc64 (pljava 1.5.5) and Debian Jessie sparc (pljava 1.5.1). With these properties pljava builds successfully in both cases.

PR #366, opened 2020-11-25
NarUtil: handle null Linker parameter to getAOL(...) as intended

angusws:
Previously we were throwing a NullPointerException if we had a null
Linker passed in, because of the line:
    if(linker != null & linker.getName() != null)

This change simply changes that single-ampersand to a double-ampersand
which is the behaviour the original author intended.


bchiodo:
Just merged. Thanks for the fix.

Issue #370, opened 2021-01-20
MSVC resource files cause the build to fail with a NullPointer exception 


vindiagram:
When using the undocumented <versionInfo> field to generate details for an executable the build fails because the .res file compilation throws a null pointer execption. Thus at the end of the build during linking the build fails because this .res file does not exist. 
This same issue occurs if a .rc file is created and included in the src/main directory.

The NullPointer exception occurs from the com.github.maven_nar.cpptasks.compiler.CommandLineCompiler function runCommand. The commands variable is null when the cmdLine arguments are added to it.

I tracked the issue to the com.github.maven_nar.NarCompileMojo in the createLibrary function when the resource compiler is added to the execution, the compilerDef commands must be set to compileCommands or an empty initialized list.

The following code block:
`
if (getResource() != null) {
  final CompilerDef res = getResource().getCompiler(Compiler.MAIN, null);
  if (res != null) {
    task.addConfiguredCompiler(res);
  }
}
`

Should be

`
if (getResource() != null) {
  final CompilerDef res = getResource().getCompiler(Compiler.MAIN, null);
  if (res != null) {
    **res.setCommands(compileCommands);**
    task.addConfiguredCompiler(res);
  }
}
`

Doing so resolves the error and allows the build to compile.

Note: I susspect this issue also exists when adding the IDL and Message compiler definitions before the resource compiler definition and thus the same line should be added to each of them as well.

PR #377, opened 2021-07-23
Fix NPE msvc .res build

GregDomjan:
The `commands` list was not initialized for resource compilation. 
Not clear on any need to do so, chosen to simply check for null and skip step of adding command


GregDomjan:
Resolved #370

Issue #371, opened 2021-01-31
Support Apple M1 chips.


syhily:
no linker is defined in the aol.properties file for 'aarch64.MacOSX.linker'

mondain:
@syhily it pretty simple to add other entries in the `aol.properties` file; I'm sure the maintainers would appreciate a patch / PR, if you or anyone else do the leg-work on this support.

mondain:
@syhily I added this section to my `aol.properties` file for M1

```
#
# MacOSX ("Mac OS X" => MacOSX) Apple Silicon M1 (64 bits)
#
aarch.MacOSX.linker=g++

aarch64.MacOSX.g++.cpp.compiler=g++
aarch64.MacOSX.g++.cpp.defines=Darwin GNU_GCC 
aarch64.MacOSX.g++.cpp.options=-Wall -Wno-long-long -Wpointer-arith -Wconversion
aarch64.MacOSX.g++.cpp.includes=**/*.cc **/*.cpp **/*.cxx
aarch64.MacOSX.g++.cpp.excludes=

aarch64.MacOSX.g++.c.compiler=gcc
aarch64.MacOSX.g++.c.defines=Darwin GNU_GCC 
aarch64.MacOSX.g++.c.options=-Wall -Wno-long-long -Wpointer-arith -Wconversion
aarch64.MacOSX.g++.c.includes=**/*.c
aarch64.MacOSX.g++.c.excludes=

aarch64.MacOSX.g++.fortran.compiler=gfortran
aarch64.MacOSX.g++.fortran.defines=Darwin GNU_GCC 
aarch64.MacOSX.g++.fortran.options=-Wall -fno-automatic -fno-second-underscore
aarch64.MacOSX.g++.fortran.includes=**/*.f **/*.for
aarch64.MacOSX.g++.fortran.excludes=

aarch64.MacOSX.g++.java.include=include
aarch64.MacOSX.g++.java.runtimeDirectory=IGNORED

aarch64.MacOSX.g++.lib.prefix=lib
aarch64.MacOSX.g++.shared.prefix=lib
aarch64.MacOSX.g++.static.extension=a
aarch64.MacOSX.g++.shared.extension=dylib
aarch64.MacOSX.g++.plugin.extension=bundle
aarch64.MacOSX.g++.jni.extension=jnilib
aarch64.MacOSX.g++.executable.extension=
```

mondain:
[...]

PR #382, opened 2021-09-21
Add support for Apple M1 build

GregDomjan:
As per discussion on #371
@syhily @mondain

### DIFF
--- a/src/main/resources/com/github/maven_nar/aol.properties
+++ b/src/main/resources/com/github/maven_nar/aol.properties
@@ -1321,3 +1321,37 @@ ppc64.AIX.gcc.plugin.extension=a
 ppc64.AIX.gcc.jni.extension=a
 ppc64.AIX.gcc.executable.extension=
 #
+
+#
+# MacOSX ("Mac OS X" => MacOSX) Apple Silicon M1 (64 bits)
+#
+aarch64.MacOSX.linker=g++
+
+aarch64.MacOSX.g++.cpp.compiler=g++
+aarch64.MacOSX.g++.cpp.defines=Darwin GNU_GCC 
+aarch64.MacOSX.g++.cpp.options=-Wall -Wno-long-long -Wpointer-arith -Wconversion
+aarch64.MacOSX.g++.cpp.includes=**/*.cc **/*.cpp **/*.cxx
+aarch64.MacOSX.g++.cpp.excludes=
+
+aarch64.MacOSX.g++.c.compiler=gcc
+aarch64.MacOSX.g++.c.defines=Darwin GNU_GCC 
+aarch64.MacOSX.g++.c.options=-Wall -Wno-long-long -Wpointer-arith -Wconversion
+aarch64.MacOSX.g++.c.includes=**/*.c
+aarch64.MacOSX.g++.c.excludes=
+
+aarch64.MacOSX.g++.fortran.compiler=gfortran
+aarch64.MacOSX.g++.fortran.defines=Darwin GNU_GCC 
+aarch64.MacOSX.g++.fortran.options=-Wall -fno-automatic -fno-second-underscore
+aarch64.MacOSX.g++.fortran.includes=**/*.f **/*.for
+aarch64.MacOSX.g++.fortran.excludes=
+
+aarch64.MacOSX.g++.java.include=include
+aarch64.MacOSX.g++.java.runtimeDirectory=IGNORED
+
+aarch64.MacOSX.g++.lib.prefix=lib
+aarch64.MacOSX.g++.shared.prefix=lib
+aarch64.MacOSX.g++.static.extension=a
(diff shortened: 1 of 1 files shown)


marschall:
I don't think this works.
1. I think you need to replace `aarch64.MacOSX.g++.*` with `aarch64.MacOSX.gpp.*`
1. I think you need to replace `aarch64.MacOSX.g++.java.include=include` with `aarch64.MacOSX.gpp.java.include=include;include/darwin`

Demogorgon314:
Any progress?

marschall:
@Demogorgon314 I have a build running with these properties 

https://github.com/marschall/exception-counter/blob/master/src/nar/apple.arm.aol.properties

GregDomjan:
@Demogorgon314 @marschall  Updated the original suggested config from @mondain based on feedback, linker name `g++` is converted to `gpp` when used to lookup properties.

I still don't have a mac to test it, but seems to be logically more on target now.

adamretter:
I would like to be able to produce a Universal Binary on macOS, see: https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary

I have an existing plugin configuration for building x86_64 on macOS x86_64 like this:
```xml
            <plugin>
                <groupId>com.github.maven-nar</groupId>
                <artifactId>nar-maven-plugin</artifactId>
                <version>3.10.2-SNAPSHOT</version>
                <extensions>true</extensions>
                <executions>
                    <execution>
                        <id>compile-native</id>
                        <phase>compile</phase>
                        <goals>
                            <goal>nar-validate</goal>
                            <goal>nar-compile</goal>
                        </goals>
                        <configuration>
                            <java>
                                <include>true</include>
                                <link>false</link>
                            </java>
                            <libraries>
                                <library>
                                    <type>executable</type>
                                    <run>false</run>
                                    <linkCPP>false</linkCPP>
                                </library>
                            </libraries>
                            <c>
                                <clearDefaultOptions>true</clearDefaultOptions>
                                <options>
                                    <option>-arch</option>
                                    <option>x86_64</option>
                                    <option>-mmacosx-version-min=10.9</option>
                                    <option>-fobjc-exceptions</option>
                                    <option>-std=c99</option>
                                    <option>-DLIBJLI_DYLIB="${java.home}/jre/lib/jli/libjli.dylib"</option>
                                    <option>-c</option> <!-- NOTE: compile only, don't link at this stage! -->
                                </options>
                            </c>
                            <linker>
                                <options>
                                    <option>-framework</option>
                                    <option>Cocoa</option>
                                    <option>-mmacosx-version-min=10.9</option>
                                </options>
                            </linker>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

I am guessing that I should be able to have two compilation executions, one for `x86_64` and one for `arm64`, and then have an additional maven plugin that just calls the `lipo` tool to combine the two binaries.

Whilst it is easy to cross-compile for arm64 macOS on an x86_64 macOS machine, I can't seem to figure out how to do it using the `nar-maven-plugin` and this PR specifically. I have to first achieve this by changing these options:
```xml
<option>-arch</option>
<option>arm64</option>
```

However I get the following errors:
```
[INFO] --- nar-maven-plugin:3.10.2-SNAPSHOT:nar-validate (compile-native) @ appbundler ---
[INFO] Using AOL: x86_64-MacOSX-gpp
[INFO]
[INFO] --- nar-maven-plugin:3.10.2-SNAPSHOT:nar-compile (compile-native) @ appbundler ---
[INFO] Compiling 1 native files
[INFO] Warning: clang cannot dynamically link libgcc
[INFO] 1 total files to be compiled.
[INFO] Found 16 processors available
[INFO] Limited used processors to 1
[INFO]
Starting Core 0 with 1 source files...
[INFO] Linking...
[INFO] OUTPUT>ld: warning: ignoring file /Users/aretter/code/evolvedbinary/appbundler-maven-build/target/nar/obj/x86_64-MacOSX-gpp-executable/main.2193ccd9.o, building for macOS-x86_64 but attempting to link with file built for unknown-arm64
[INFO] OUTPUT>Undefined symbols for architecture x86_64:
[INFO] OUTPUT>  "_main", referenced from:
[INFO] OUTPUT>     implicit entry/start for main executable
[INFO] OUTPUT>ld: symbol(s) not found for architecture x86_64
[ERROR] OUTPUT>clang: error: linker command failed with exit code 1 (use -v to see invocation)
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

However running file shows that an arm64 binary is produced, i.e.:
```
$ file target/nar/obj/x86_64-MacOSX-gpp-executable/main.2193ccd9.o

target/nar/obj/x86_64-MacOSX-gpp-executable/main.2193ccd9.o: Mach-O 64-bit object arm64
```

Any ideas?

adamretter (changes requested):
Some small changes are needed in practice

PR #394, opened 2022-12-19
Handle NPEs better - config errors logged as Warn; mitigate usage errors

susanw1:
1. Plugin fails with NPE when user provides empty definitions eg &lt;define>&lt;/define> - which can happen innocently when a Maven property is empty (eg &lt;define>${foo}&lt;/define>). I've changed it to log a WARN now, because it's probably caused by a config error (it was in my case!), and the user probably wants to know about it. I've factored out the '='-splitting code, because we do that in a few cases (not all of them can fail this way, but I think it's better to do the parsing in one place). 
2. In CUtil.runCommand, it was assumed that the Environment env is non-null when it extracts the Variables vector, but the various usages presume it can be null, resulting in NPE. I've softened it up to produce an empty vector instead.
3. Updated deprecated calls to StringUtil.replace.

### DIFF
--- a/src/main/java/com/github/maven_nar/Compiler.java
+++ b/src/main/java/com/github/maven_nar/Compiler.java
@@ -263,7 +263,6 @@ public abstract class Compiler {
       this.mojo.getLog().warn("NAR configuration: empty " + configName + " - ignoring");
       return null;
     }
-    final DefineArgument define = new DefineArgument();
     final String[] pair = argument.split("=", 2);
 
     if (pair[0].equals("")) {
@@ -271,6 +270,7 @@ public abstract class Compiler {
         return null;
     }
 
+    final DefineArgument define = new DefineArgument();
     define.setName(pair[0]);
     define.setValue(pair.length > 1 ? cleanDefineValue(pair[1]) : null);
     return define;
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/maven_nar/Compiler.java
```diff
@@ -250,6 +250,32 @@ public abstract class Compiler {
...
+   * @param argument the string to parse (may be null)
+   * @param configName the name of the argument type, for logging purposes
+   * @return the parsed DefineArgument, or null if argument is null
+   */
+  private DefineArgument parseDefineArgument(final String argument, String configName) {
+    if (argument == null) {
+      this.mojo.getLog().warn("NAR configuration: empty " + configName + " - ignoring");
+      return null;
+    }
+    final DefineArgument define = new DefineArgument();
```
GregDomjan, line 266: could declare this after the following conditional that is returning null

GregDomjan:
Seems like fair refactoring and improvement to cover NPE

src/main/java/com/github/maven_nar/Compiler.java
susanw1: Good point - moved. Thanks!

